PR #3, opened 2015-10-12
Fix issue regarding Char type, it was handled as Object instead of St…

yovio:
For property with Char type, it was handled as Object.

I guess it should be handles as String.


rsuter:
Thanks for fixing this..

rsuter:
Fix is now available via NuGet.

yovio:
Thank you, I can verify that its available in Nuget now.

PR #8, opened 2015-12-10
Type to schema fixes

JoshuaWhatley:
Fixed issues:
 - The conversion of types inheriting from Dictionary were failing
 - The generation of non-generic collections (ex: JObject) were failing

### DIFF
--- a/src/NJsonSchema.Tests/Conversion/TypeToSchemaTests.cs
+++ b/src/NJsonSchema.Tests/Conversion/TypeToSchemaTests.cs
@@ -191,18 +191,6 @@ namespace NJsonSchema.Tests.Conversion
         }
 
         [TestMethod]
-        public void When_converting_non_generic_collection_then_items_must_correctly_be_loaded()
-        {
-            //// Act
-            var schema = JsonSchema4.FromType<JObject>();
-
-            //// Assert
-            Assert.AreEqual(JsonObjectType.Object, schema.Type);
-            Assert.AreEqual(typeof(JObject).Name, schema.TypeName);
-        }
-
-
-        [TestMethod]
         public void When_converting_array_then_items_must_correctly_be_loaded()
         {
             When_converting_array_then_items_must_correctly_be_loaded("Array");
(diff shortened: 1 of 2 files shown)


rsuter:
Why would you generate the JSON Schema for JObject? What do you expect?

JoshuaWhatley:
We are using NJsonSchema in conjunction with Swagger (http://swagger.io/) to help present the request / response models of some of our APIs. JObject happens to be present in one of these models.

rsuter:
Just FYI: NJsonSchema is used in http://nswag.org a .NET Swagger toolchain

JoshuaWhatley:
Cool, I wasn't aware of that project. Definitely useful tools, thanks!

rsuter:
Ok, but JObject is a complex object and not serialized as JObject but as its data...

JoshuaWhatley:
I understand. I'll address the JObject scenario before it reaches NJsonSchema. I think it makes more sense that way. I'll remove that test case and the modifications to the array flow.

rsuter:
The other problem can be fixed by uncommenting these lines: 

https://github.com/NJsonSchema/NJsonSchema/blob/master/src/NJsonSchema/JsonObjectTypeDescription.cs#L98
https://github.com/NJsonSchema/NJsonSchema/blob/master/src/NJsonSchema/JsonObjectTypeDescription.cs#L108

The test is here: 
https://github.com/NJsonSchema/NJsonSchema/blob/master/src/NJsonSchema.Tests/Conversion/ArrayTypeToSchemaTests.cs#L11

The problem is that with this enabled, only dictionaries and array types from System.* are directly treated as JSON array, the rest are handled with inheritance... If someone has an own list implementation, the class is generated as own class but not as JSON array...

rsuter:
There is not an easy solution and your change may break other scenarios...

rsuter:
JObject handling is now implemented: 
https://github.com/NJsonSchema/NJsonSchema/blob/master/src/NJsonSchema.Tests/Conversion/TypeToSchemaTests.cs#L187

JoshuaWhatley:
Thanks man, appreciate it.

rsuter:
Ok, both features are now available... see https://github.com/NJsonSchema/NJsonSchema/commit/4572026b2e639ddce7293008b13b1b43111ec8c2

PR #24, opened 2016-02-12
Updating File.txt to add using statements for collections

prateekguru:
Updating File.txt to add using statements for collections (ObservableCollection was missing the required using statement)

### DIFF
--- a/src/NJsonSchema.CodeGeneration/CSharp/Templates/File.txt
+++ b/src/NJsonSchema.CodeGeneration/CSharp/Templates/File.txt
@@ -1,5 +1,9 @@
-﻿using System;
+using System;
+using System.Collections.ObjectModel;
 using System.ComponentModel;
+using System.Collections.Specialized;
+using System.Collections;
+using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using Newtonsoft.Json;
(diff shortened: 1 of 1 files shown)


rsuter:
Is this still working with all .NET profiles (i.e. PCL, UWP, etc)?

rsuter:
Otherwise you have to add the namespace manually with the `AdditionalNamespaceUsages` setting.

prateekguru:
I haven't tested it outside of .NET 45.
Also, I cannot find AdditionalNamespaceUsages; Where is that setting?

prateekguru:
I am looking at the Code Generation section primarily.

prateekguru:
Also, the namespaces, I have added to File.txt are available in all versions of .NET that are supported currently. (UWP, Silverlight, .NET 4x and PCL)

rsuter:
Hmm i think this setting is missing here... But you can use the full qualified type name and it should work without additional usages

prateekguru:
Yes, can do that. 
Thanks

rsuter:
But i will think about adding these namespaces..

rsuter:
The namespaces should be available in PCL/all target frameworks... Will be available in the next nuget version.

PR #35, opened 2016-03-09
Base64 string validation

Samuron:
Added validation of string properties with format set to Base64

### DIFF
--- a/src/NJsonSchema.Tests/NJsonSchema.Tests.csproj
+++ b/src/NJsonSchema.Tests/NJsonSchema.Tests.csproj
@@ -60,6 +60,7 @@
     <Compile Include="Conversion\SchemaReferenceTests.cs" />
     <Compile Include="Conversion\ArrayTypeToSchemaTests.cs" />
     <Compile Include="Generation\AnnotationsGenerationTests.cs" />
+    <Compile Include="Validation\FormatBase64Tests.cs" />
     <Compile Include="Validation\FormatIpV6Tests.cs" />
     <Compile Include="Generation\AttributeGenerationTests.cs" />
     <Compile Include="Generation\EnumGenerationTests.cs" />
--- a/src/NJsonSchema.Tests/Validation/FormatBase64Tests.cs
+++ b/src/NJsonSchema.Tests/Validation/FormatBase64Tests.cs
@@ -1,0 +1,72 @@
+﻿using System;
+using System.Linq;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json.Linq;
+using NJsonSchema.Validation;
+
+namespace NJsonSchema.Tests.Validation
+{
+    [TestClass]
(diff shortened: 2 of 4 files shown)


rsuter:
Thanks for contributing

rsuter:
Maybe we should also support `byte` which looks like it is the same as `base64`, see https://developers.google.com/discovery/v1/type-format#type-and-format-summary ?
Is there a specification what is correct?

rsuter:
The official spec does not list either `byte` or `base64`: 
http://json-schema.org/latest/json-schema-validation.html#anchor104

Samuron:
Okay, there is definitely an inconsistency in behavior: generating schema from CLR type with property of `byte[]` emits `base64` format. Which is not a part of the standard. 
What about leveraging `media` keyword as in https://github.com/json-schema/json-schema/wiki/Media? It has not as strict validation semantics: 

> Implementations MAY, but are not required to, validate a JSON instance against this keyword

So, perhaps, the strictness should be configurable.
Or use `byte` keyword as it is already used despite of official standard:
https://github.com/json-schema/json-schema/wiki/%22format%22-suggestions

rsuter:
Yes, but `media` is also just a proposal (and currently not part of NJsonSchema)...

rsuter:
The `byte[]` to `base64` conversion can be found in the JsonObjectTypeDescription class: https://github.com/NJsonSchema/NJsonSchema/blob/master/src/NJsonSchema/JsonObjectTypeDescription.cs#L59

According to https://github.com/json-schema/json-schema/wiki/%22format%22-suggestions maybe we should use `byte` instead of `base64` everywhere... or support the `media` property...

rsuter:
... do you need a new NuGet package version with this PR immediately?

Samuron:
I think that using `byte` Instead of `base64` is a good idea, as it is already used by some other APIs, according to the links. Despite not being defined in the standard, it seems that such validation provides much more safety in terms of deserializing to CLR types. Let me do it, I will try to send a PR for review soon.

rsuter:
Also, in [the article](https://github.com/json-schema/json-schema/wiki/%22format%22-suggestions) I read that the Swagger spec uses byte instead of base64 and I mainly use the NJsonSchema library in the [NSwag](http://nswag.org) Swagger toolset for .NET project. So it is a good fit...

rsuter:
So, your idea is to completely remove `base64` format and use `byte` instead? Maybe you should also enhance the XML doc of the `byte` enum with a note that the format is base64

rsuter:
Maybe we should support `byte` and `base64` for validation and use `byte` for JSON Schema generation (from C# types).

Samuron:
Actually I think it is better to support both, so the schemes generated with previous version of the lib would still be recognized, but start to emit `byte`.

rsuter:
Yes, these things have to be planned carefully because we do not want to break existing code...

PR #36, opened 2016-03-10
Obsoleting Base64 format

Samuron:
According to our discussion made `base64` format obsolete and replaced it with `byte`.


Samuron:
It seems that CI have failed due to usages of obsolete. I have used pragmas where had compiler warning, but it seems i've missed something.

PR #56, opened 2016-04-21
Pattern properties problems fix

pmezykowski:
fixed problems with validating pattern properties + some test added to be sure that it really works and will not happen again :)

I've added just basic unit tests for that, but there could be many more of them to fully test the functionality :)


rsuter:
Thanks for fixing this issue.. Im always looking for help :-)

PR #72, opened 2016-05-13
TypeRaw property changed to deserialize without using JsonConvert.DeserializeObject

harold-b:
Deserializing the `TypeRaw` property choses from the 7 allowed primitive types directly, instead of using DeserializeObject.

Thanks for the lib :+1:


rsuter:
Thanks for the pr and using the lib

PR #81, opened 2016-05-24
Locking only in the path that modifies the cache.

jkonecki:
I removed the thread locking that was turing the cache into single-threaded. Locking is done only when the cache is modified.

### DIFF
--- a/src/NJsonSchema/Infrastructure/ReflectionCache.cs
+++ b/src/NJsonSchema/Infrastructure/ReflectionCache.cs
@@ -17,64 +17,71 @@ namespace NJsonSchema.Infrastructure
 {
     internal static class ReflectionCache
     {
-        private static readonly object Lock = new object();
-
         private static readonly Dictionary<Type, IList<PropertyInfo>> PropertyCacheByType = new Dictionary<Type, IList<PropertyInfo>>();
         private static readonly Dictionary<PropertyInfo, CustomAttributes> AttributeCacheByProperty = new Dictionary<PropertyInfo, CustomAttributes>();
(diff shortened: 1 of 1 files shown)


rsuter:
Maybe we should implement proper double-checked-locking then...

rsuter:
I.e.: 
```
if (!dictionary.ContainsKey(key))
{
	lock (dictionary)
	{
		if (!dictionary.ContainsKey(key))
		{
			...
		}
	}
}
return dictionary[key];
```
But I'm not sure if `ContainsKey` and `dictionary[key]` is 100% thread-safe...

rsuter:
See http://stackoverflow.com/questions/6018768/double-checked-locking-on-dictionary-containskey

rsuter:
I think there are two options: 

1. Lock whole methods (as is) but with one lock per method (i.e. use dictionary as locks)
2. Use ConcurrentDictionary and double-checked locking (but the current PCL does not support this)

I dont think 1. will degrade performance that much... as soon as a key is loaded its extremely fast (acquiring lock & get existing value)

jkonecki:
I'm confused. I've actually implemented double checked locking in this PR....

rsuter:
The problem is, that in double-checked locking you check outside the lock and inside the lock again and the check (ie. expression in the if) and access must be atomic which `ContainsKey` and `dictionary[key]` aren't... (see the stackoverflow question)

jkonecki:
Ah yes. :-( 

In that case, since we cannot use Concurrent Dictionary, locking on the while method is the way to go. I'll modify PR (will keep 3 separate locks). 
Won't be much of a performance impact since we are single threaded at the moment.

PR #90, opened 2016-05-31
Cs fallback type name from schema

lheiskan:
Relates to: https://github.com/NJsonSchema/NJsonSchema/issues/89


rsuter:
Thanks for the PR, but this has to be integrated into much more classes than only the generator itself... Ill do this as soon as possible..

rsuter:
This can now be implemented using the ITypeNameGenerator setting

PR #100, opened 2016-06-05
Array ref support

lheiskan:
This test surfaces an issue where Anonymous types are generated for a ref type that is well defined.

i.e. if a type ref is used, i would expect the ref type be generated as it's own class and the generated code should contain a typed collection of type ref type.


lheiskan:
Added some code that adds support for generating the type names, tests now pass. The code parses a typename hint from json path used in the ref, and places the type hint into schema.RawTypeName it if is null. From there, code works the same way as if the x-typeName property had been specified in the file.

Issue #11, opened 2015-12-17
Support allOf with code generator


rsuter:
Sample JSON (single inheritance, properties defined in allOf): 

    "Pet": {
      "type": "object",
      "allOf": [
        {
          "$ref": "#/definitions/NewPet"
        },
        {
          "required": [
            "id"
          ],
          "properties": {
            "id": {
              "type": "integer",
              "format": "int64"
            }
          }
        }
      ]
    },
    "NewPet": {
      "type": "object",
      "required": [
        "name"
      ],
      "properties": {
        "name": {
          "type": "string"
        },
        "tag": {
          "type": "string"
        }
      }
    },

Should generate in TypeScript: 

	export interface Pet implements NewPet {
             id: integer; 
	}

	export interface NewPet {
		name: string;
		tag?: string;
	}

PR #104, opened 2016-06-06
allOf support (simple)

lheiskan:
Hey,

Ended up adding a simple allOf support for CS codegen. The idea is to expand any allOf definitions into the aggregate type API. 

This could probably be developed further by generating interfaces for the types in the aggregates, and having the aggregate type implement those interfaces. It also might be nice if aggregate type would actually wrap real instances of the types making up the aggregate - and provide ways for constructing the aggregate using these instances. 

However, i'll experiment with this as the first step to get some code generated for the aggregates. Seems like the api resulting from this simple implementation should not break even more features like the above are added later.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.Tests/CSharp/CSharpGeneratorTests.cs
+++ b/src/NJsonSchema.CodeGeneration.Tests/CSharp/CSharpGeneratorTests.cs
@@ -8,6 +8,54 @@ namespace NJsonSchema.CodeGeneration.Tests.CSharp
     [TestClass]
     public class CSharpGeneratorTests
     {
+        [TestMethod]
+        public void multiple_refs_in_all_of_should_expand_to_single_def()
+        {
+            var schema = @"{
+                '$schema': 'http://json-schema.org/draft-04/schema#',
+                'id': 'http://some.domain.com/foo.json',
+                'type': 'object',
+                'additionalProperties': false,
+                'definitions': {
+                    'tRef1': {
+                        'type': 'object',
+                        'properties': {
+                            'val1': {
+                                'type': 'string',
+                            }
+                        }
+                    },
+                    'tRef2': {
+                        'type': 'object',
+                        'properties': {
+                            'val2': {
+                                'type': 'string',
+                            }
+                        }
+                    },
+                    'tRef3': {
+                        'type': 'object',
+                        'properties': {
+                            'val3': {
+                                'type': 'string',
+                            }
+                        }
+                    }
+                },
+                'tAgg': {
+                    'allOf': [
+                        {'$ref': '#/definitions/tRef1'},
+                        {'$ref': '#/definitions/tRef2'},
+                        {'$ref': '#/definitions/tRef3'}
+                    ]
+                }
+            }";
+            var s = NJsonSchema.JsonSchema4.FromJson(schema);
+            var settings = new CSharpGeneratorSettings() { ClassStyle = CSharpClassStyle.Poco, Namespace = "ns", };
+            var gen = new CSharpGenerator(s, settings);
+            var output = gen.GenerateFile();
+            System.Console.WriteLine(output);
+        } 
 
         [TestMethod]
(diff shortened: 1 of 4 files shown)


rsuter:
I think this will give lots of trouble, because allOf is used for inheritance... this way you will also render the properties of the inherited class...

lheiskan:
I guess the question is how to model cases that would require multiple inheritance in languages that don't support it - not sure if this is actually the way to go, however, it seems somewhat useful - it basically does what i would otherwise need to do manually.

It would probably be a good idea to support a single inheritance in c# as a special case though. And probably with more effort, the multiple inheritance could be modeled more elegantly (i.e. via interfaces and wrappers or something like that).

rsuter:
So the first oneOf is inheritance and the rest is merged? Or only the oneOf with ref? Its a little too much magic for me.. Cant you genereate the properties directly in the schema?

rsuter:
See https://github.com/NJsonSchema/NJsonSchema/issues/11

rsuter:
With your PR, you would kill the inheritance feature because all referenced schemas in allOf are merged together...

lheiskan:
hey,

thanks for the comments, this implementation certainly is naive, admitted :). 

It seems at least somewhat backward compatible as the test that does single inheritance still works (When_allOf_contains_one_schema_then_csharp_inheritance_is_generated) - so not sure it this would actually kill the inheritance feature - least i was hoping that it would not break it.

#11 works well for typescript, and I think we could do something similar on c# side as well. On c# side  we will probably need to generate interfaces and the actual types. Aggregate types could then wrap instances of the sub components and implement all sub component interfaces. I was hoping that the expansion of properties would be compatible with something like this, as I surely think that it bare expansion is a primitive solution.

One interesting question is also do we want to distinguish between single inheritance and multiple inheritance scenarios - or should both be handled in the same way (which would mean no-inheritance in c# based solution, as there is no multiple inheritance available).

Not sure what you mean by adding properties directly into the schema? Do you mean modifying the input schema and converting any allOf items to a flat list of properties?

rsuter:
I think When_allOf_contains_one_schema_then_csharp_inheritance_is_generated works because it only tests the inheritance but the properties are probably generated in the class and the base class...

rsuter:
> Not sure what you mean by adding properties directly into the schema? Do you mean modifying the input schema and converting any allOf items to a flat list of properties?

Yes

lheiskan:
Hey, 

modified the branch so that the expansion magic kicks in when single inheritance cannot help - a bug by my part, good call. 

In my case I'm using a schema that's maintained by another party, modifications can be a hard sell. I suspect this situation is quite common - there is always a workaround though, which is to write the wrapper code by hand, in which case my first version would look pretty much the same as is being generated here, although, doing this manually would allow also for more elaborate support. So perhaps, if this thing goes in, a setting that would allow to opt in for this would be in order.

rsuter:
Are these rules ok for you?
https://github.com/NJsonSchema/NJsonSchema/wiki/JsonSchema4#inheritance-and-allof-handling

rsuter:
I copied your unit test but removed the lines "type": "object" because it would be an inheritance

rsuter:
See https://github.com/NJsonSchema/NJsonSchema/commit/ae8d518c8f2fdaf860a2b591862bf642eb0d4b15

lheiskan:
Hey!

I'm not sure if it is necessary to make a further distinction regarding inheritance in the schema level model - as allOf has quite clear semantics. Would it be cleaner if codegen step would receive such additonal hints via it's settings, allowing users to select how to handle allOf?

rsuter:
Do you really need this? All these features make the library more complicated.. :-)

lheiskan:
Hi,

I can certainly live with the current state of affairs (i.e. i have my own branch anyway) - and I agree that keeping the code simple is very important. The merge all props solution is also an interim/workaround level solution, so now may not be the right time to bring this in. Totally your call.

I guess users of the library will run into these constructs quite often - so of there is more pressure supporting allOf we can always come back to this.

PR #107, opened 2016-06-07
Test to uncover a potential issue in allOf inheritance generation

lheiskan:
i.e. when schema contains two items in allOf array, then aggregate type should
be created via inheritance.

ran into this when playing around with allOf, looks like it could be an issue.


rsuter:
Can you retest this with v2.39?

rsuter:
This unit test is green in the current version...

PR #108, opened 2016-06-07
Array empty schema support

lheiskan:
Here's a test for a potential issue - test crashes with a null ref exception for a schema that contains a field of type array, and has no items property.

Json schema specification docs indicate that schema is valid and should be interpreted as if items property was specified with an empty schema as it's value.

Validation keywords for arrays: http://json-schema.org/latest/json-schema-validation.html#anchor36


rsuter:
According to the spec: "The value of "items" MUST be either an object or an array". So null should throw an exception because schema is not valid...

lheiskan:
Sure - however, the next section indicates that if items is missing the validation should always succeed. And the default values section that follows indicates what to do if they are missing. So seems to me that the first section should be interpreted as 'if items is specified, then it's value must be...'. 

Ran into one such file in the wild, so it seems like they exist and seem to pass at least some validators.

lheiskan:
btw. i still have the test, just dropped the branch as figured it would be easier to merge without the extra crud.

PR #109, opened 2016-06-07
Test uncovering an issue with arrays with no items handling

lheiskan:
According to json schema specs schema with an array type, but no items, are
valid. They should be interpreted as if the items was present and contained an
empty schema value.

http://json-schema.org/latest/json-schema-validation.html#anchor36


rsuter:
Are you sure that this is allowed. The spec is not very clear here...

lheiskan:
At least some validators will consider the following valid
```json
{
	'properties': {
    	'emptySchema': { 'type': 'array' }
     }
}
```
for instance this one: http://www.jsonschemavalidator.net/

rsuter:
OK, we can implement this as fallback...

rsuter:
Fixed in https://github.com/NJsonSchema/NJsonSchema/commit/4421ec567d1cd8b1f3b37b1954a5d0372482d896

Issue #115, opened 2016-06-10
Support for handling different Property Name conventions


jkonecki:
When using Json.NET `CamelCasePropertyNamesContractResolver` the serialized data has camel case property names. However NJsonSchema generates upper-case property names (based on reflection).

It would be beneficial to allow users to control the naming convention of the generated property names.

Suggested solution:
1. Define enum PropertyNameHandling { Default, CamelCase }
2. Add a new option `DefaultPropertyNameHandling` to `JsonSchemaGeneratorSettings`
3. Pass the naming convention to `Property.GetName()`

jkonecki:
Should the Default `PropertyNameHandling` been actually called Reflection?

rsuter:
CamelCasePropertyNamesContractResolver   inherits from DefaultContractResolver. So I think the DefaultContractResolver implements this reflection and thus Default would be ok?

rsuter:
The question is whether CamelCasePropertyNamesContractResolver  applies ToCamelCase to the property name or also to the reflected name (respecting JsonProperty and DataMember attributes)

jkonecki:
I'll double-check and write appropriate tests.

rsuter:
Can we close this issue?

jkonecki:
Yes. I'm happy with the implementation.

PR #117, opened 2016-06-10
Camel-case property name handling

jkonecki:
Implementation of issue #115


rsuter:
Why dont we use `return new CamelCasePropertyNamesContractResolver().GetResolvedPropertyName(property.Name);` insead of your `ToCamelCase`? This way the conversion is always implemented as expected...

jkonecki:
Good idea. I didn't know you would be happy to have an explicit dependency on this class. I'll amend the PR shortly and update the comment for enum value

rsuter:
We already have a dependency to Json.NET so its fine for me...

PR #121, opened 2016-06-14
Workaround for .NET portable MemberInfo comparisons failing with derived classes.

lyntondicks:
JsonPathUtilities.GetPropertyName failing when property is defined in base class, but reflected from derived class. See: https://github.com/dotnet/corefx/issues/5884

I ran into this problem when using NSwag with [FromUri] parameter that was a derived class.


rsuter:
The problem here is that all attributes (e.g. DataMember and JsonProperty attributes) are completely ignored... Isn't there a way to read them or better, to fix this so that the behavior is correct?

rsuter:
Why not use

    return ReflectionCache.GetProperties(property.DeclaringType).First(p => p.PropertyInfo.Name == property.Name).GetName();

instead of 

    return ReflectionCache.GetProperties(property.DeclaringType).First(p => p.PropertyInfo == property).GetName();

rsuter:
I implemented this in v2.44, please try with this version... if there's a problem with that, please make a suggestion..

lyntondicks:
Thanks, looks like it's working correctly.

PR #134, opened 2016-06-28
Added support for specifying explicit enumName values using DisplayAtttribute on enum members

richhe:
The current implementation for generating enumNames in a schema uses the enum member name which is constrained to being a valid c# identifier. This change enables the value used for the enumName to be specified by applying a System.ComponentModel.DisplayAttribute to the original enum which provides the flexibility to include spaces (or other extended characters) in an enumName as per the example in the draft proposal at https://github.com/json-schema/json-schema/wiki/enumNames-(v5-proposal)

### DIFF
--- a/src/NJsonSchema.Tests/Generation/EnumGenerationTests.cs
+++ b/src/NJsonSchema.Tests/Generation/EnumGenerationTests.cs
@@ -3,6 +3,8 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
 using NJsonSchema.Generation;
+using System.Runtime.Serialization;
+using System.ComponentModel.DataAnnotations;
 
 namespace NJsonSchema.Tests.Generation
 {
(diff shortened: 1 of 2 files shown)


rsuter:
I think using the DisplayAttribute is wrong, because the enum names specify an identifier (ie a human readable ID), but the DisplayAttribute specifies a display name (i.e. translated, not unique, etc.). I heavily rely on the enumNames for code generation and this change may introduce breaking changes in existing scenarios... My suggestion is to implement a custom attribute for this. Or is there a non-display based attribute (ie. `NameAttribute`)

richhe:
Thanks Rico, I can't find an existing attribute that's a good fit so will update to add a custom NameAttribute as you suggest. Many thanks

rsuter:
Maybe JsonNameAttribute or JsonSchemaNameAttribute or NameAttribute (name clash?)?

Add an attribute class here: 
https://github.com/NJsonSchema/NJsonSchema/tree/master/src/NJsonSchema/Annotations

gzak:
I think DisplayName is pretty standard for this sort of this... Doesn't swagger itself rely on DisplayName to render enum values in the generated documentation?

RSuter:
It uses EnumValueAttribute - DisplayNameAttribute is a human readable description... but not a human readable ID (e.g. "The person name" vs. "PersonName")

RicoSuter:
Completely outdated - please open another PR if still a problem.

PR #146, opened 2016-07-12
Make explicit what are the additional properties causing the validation failure

rogerbramon:
Currently, when no additional properties are allowed the validation error give you no clue about what properties are causing the failure. 

With this PR I propose to add an error per property, so that we can easily know where is the problem.  

For example, giving the following schema,

```json
{
 "type": "object",
 "properties": {
   "foo": {
      "type": "object",
      "additionalProperties": false
   }
 },
 "additionalProperties": false
}
```

when validating this JSON,

````json
{ "foo": { "bar": "mybar" } }
```

we will get the following ValidationError:

```csharp
{NoAdditionalPropertiesAllowed: #/foo.bar}
    Kind: NoAdditionalPropertiesAllowed
    Path: "#/foo.bar"
    Property: "bar"
```

We are really interested in having this information when running validations in our CI server. 

Looking forward your feedback.

Thanks,

Roger

### DIFF
--- a/src/NJsonSchema.Tests/Validation/PatternPropertyValidationTests.cs
+++ b/src/NJsonSchema.Tests/Validation/PatternPropertyValidationTests.cs
@@ -25,8 +25,11 @@ namespace NJsonSchema.Tests.Validation
             var errors = schema.Validate(token);
 
             //// Assert
-            Assert.AreEqual(1, errors.Count());
-            Assert.AreEqual(ValidationErrorKind.NoAdditionalPropertiesAllowed, errors.First().Kind);
+            Assert.AreEqual(2, errors.Count());
+            foreach (var validationError in errors)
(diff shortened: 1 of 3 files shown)


rsuter:
Its a good idea to privide more validation info.. this may be a breaking change, but i think we can live with that..

rsuter:
Thanks for the pr and using the project..

rsuter:
New version will be available soon..

rogerbramon:
Very much appreciated. Looking forward the new version. BTW, I'm working on another PR to provide ToString methods for the both ValidationError and ChildSchemaValidationError. Hopefully it will be ready tomorrow.

Regards.

rsuter:
Check v2.57

rsuter:
Check v2.57

PR #148, opened 2016-07-13
Add ToString method to validation error objects

rogerbramon:
In order to improve error reporting, in this PR I propose to add ToString methods for both ValidationError and ChildSchemaValidationError.

For instance, given the following schema:

```json
{
  "type": "object",
  "properties": {
    "data": {
      "oneOf": [
        {
          "$ref": "#/definitions/DataModel"
        },
        {
          "type": "null"
        }
      ]
    }
  },
  "definitions": {
    "DataModel": {
      "type": "object",
      "properties": {
        "foo": {
          "type": "integer"
        }        
      }
    }
  }
}
```
when validating this JSON,
```json
{ "data": { "bar": 4 } }
```
we would get this string error:
```
NotOneOf error in #/data:
{
  NoAdditionalPropertiesAllowed: #/data.bar
}
{
  NullExpected: #
}
```

Looking forward your feedback.

Roger

### DIFF
--- a/src/NJsonSchema/Validation/ChildSchemaValidationError.cs
+++ b/src/NJsonSchema/Validation/ChildSchemaValidationError.cs
@@ -25,6 +25,24 @@ namespace NJsonSchema.Validation
         }
         
         /// <summary>Gets the errors for each validated subschema. </summary>
-        public IReadOnlyDictionary<JsonSchema4, ICollection<ValidationError>> Errors { get; private set; } 
+        public IReadOnlyDictionary<JsonSchema4, ICollection<ValidationError>> Errors { get; private set; }
+
+        /// <summary>Returns a string that represents the current object.</summary>
+        /// <returns>A string that represents the current object.</returns>
+        /// <filterpriority>2</filterpriority>
+        public override string ToString()
+        {
+            var output = string.Format("{0} error in {1}:\n", Kind, Path);
+            foreach (var error in Errors)
+            {
+                output += "{\n";
+                foreach (var validationError in error.Value)
+                {
+                    output += string.Format("  {0}\n", validationError.ToString().Replace("\n", "\n  "));
+                }
+                output += "}\n";
+            }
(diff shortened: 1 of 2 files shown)


rsuter:
I think the structure is not optimal, but I don't have a better idea... I'd also use "error: path" in the ToString method of ChildSchemaValidationError, ie

	NotOneOf: #/data
	{
	    NoAdditionalPropertiesAllowed: #/data.foo
	    NoAdditionalPropertiesAllowed: #/data.bar
	}
	{
	    NullExpected: #
	}

rsuter:
Mustn't it be 

    NullExpected: #/data

?
Is this a bug?

rogerbramon:
Regarding structure, I agree it's not optimal. I was testing different options and asking for opinion at work and we came up with this proposal. 

About the mentioned bug, I was about to discuss with you some potential miss-alignments when creating validation errors basically related with the path, ie

Taking into account the schema mentioned before, when validating this JSON

```json
{"data": {"foo": "not_integer", "bar": 4} }
```
we get this error,

```
NotOneOf: #/data
{
  IntegerExpected: #/foo
  NoAdditionalPropertiesAllowed: #/data.bar
}
{
  NullExpected: #
}
```

rsuter:
Thanks for the PR, the structure is fine for me... it seems that the paths are not always created correctly... i currently dont have the time to fix that.. can you create an issue for that so that we can fix it later? Or a pr which actually fixes that?

rogerbramon:
So, just to make sure, would you expect to get the error with absolute paths like this?

```
NotOneOf: #/data
{
  IntegerExpected: #/data.foo
  NoAdditionalPropertiesAllowed: #/data.bar
}
{
  NullExpected: #/data
}
```

rsuter:
Yes, you too?

rsuter:
# means document root...

rogerbramon:
yes, me too :)

PR #172, opened 2016-08-10
Only put required annotations on required properties

tstrass:
This makes it so that the C# code generator only puts required-property annotations on the properties in the `required` array in the JsonSchema.
  - It was putting the `Required` property on every `JsonProperty`, regardless of if it was actually required

    ```c#
    [JsonProperty("<#=property.Name#>", Required = <#=property.JsonPropertyRequired#>)]
    ```
  - It was also putting the `[Required]` annotation on every property of type `string`, `object`, or `array`

### DIFF
--- a/src/NJsonSchema.CodeGeneration.Tests/CSharp/CSharpGeneratorTests.cs
+++ b/src/NJsonSchema.CodeGeneration.Tests/CSharp/CSharpGeneratorTests.cs
@@ -446,7 +446,7 @@ namespace NJsonSchema.CodeGeneration.Tests.CSharp
             var output = generator.GenerateFile();
 
             // Assert
-            Assert.IsTrue(output.Contains(@"[JsonProperty(""foo-bar"", "));
+            Assert.IsTrue(output.Contains(@"[JsonProperty(""foo-bar"""));
             Assert.IsTrue(output.Contains(@"public string FooBar"));
         }
 
@@ -609,7 +609,7 @@ namespace NJsonSchema.CodeGeneration.Tests.CSharp
             var code = generator.GenerateFile();
 
             //// Assert
-            Assert.IsTrue(code.Contains("[JsonProperty(\"Foo\", Required = Required.DisallowNull)]"));
+            Assert.IsTrue(code.Contains("[JsonProperty(\"Foo\")]"));
             Assert.IsTrue(code.Contains("public string Foo1 { get; set; }"));
         }
 
--- a/src/NJsonSchema.CodeGeneration/CSharp/Models/PropertyModel.cs
+++ b/src/NJsonSchema.CodeGeneration/CSharp/Models/PropertyModel.cs
@@ -38,6 +38,8 @@ namespace NJsonSchema.CodeGeneration.CSharp.Models
 
         public string FieldName => ConversionUtilities.ConvertToLowerCamelCase(GetGeneratedPropertyName(), true);
 
+        public bool IsRequired => _property.IsRequired;
+
         public string JsonPropertyRequired
         {
             get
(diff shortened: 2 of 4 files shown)


rsuter:
Thanks for this PR, I have to check this first, I think the current implementation is fine, because I worked lots of time on the required/nullable feature of JSON Schema and Swagger...

Very important: JSON Schema property required does not state if a property is nullable - it defines if the property must be specified in the JSON data (but it can still be null). This is why I think your change is wrong...

tstrass:
Maybe I am misunderstanding you, but I don't think your comment is at odds with my changes. I'll try to be more clear with an example:

As it stands, this JSON schema
```JSON
{
  "type": "object",
  "typeName": "Person",
  "required": [
    "FirstName",
    "LastName"
  ],
  "properties": {
    "FirstName": {
      "type": "string"
    },
    "MiddleName": {
      "type": "string"
    },
    "LastName": {
      "type": "string"
    }
  }
}
```

Generates this C#:
```C#
[JsonObject(MemberSerialization.OptIn)]
[GeneratedCode("NJsonSchema", "3.4.6065.33501")]
public partial class Person : INotifyPropertyChanged
{ 
    private string _firstName; 
    private string _middleName; 
    private string _lastName;
    
    [JsonProperty("FirstName", Required = Required.Always)]
    [Required]
    public string FirstName
    {
        get { return _firstName; }
        set 
        {
            if (_firstName != value)
            {
                _firstName = value; 
                RaisePropertyChanged();
            }
        }
    }
    
    [JsonProperty("MiddleName", Required = Required.DisallowNull)]
    [Required]
    public string MiddleName
    {
        get { return _middleName; }
        set 
        {
            if (_middleName != value)
            {
                _middleName = value; 
                RaisePropertyChanged();
            }
        }
    }
    
    [JsonProperty("LastName", Required = Required.Always)]
    [Required]
    public string LastName
    {
        get { return _lastName; }
        set 
        {
            if (_lastName != value)
            {
                _lastName = value; 
                RaisePropertyChanged();
            }
        }
    }
}
```

The issue here as I see it is twofold:
- `MiddleName` gets the `[Required]` annotation, but it is not required in the JSON schema
- Even if you fix that and remove `[Required]`, there is still the other annotation, `[JsonProperty("MiddleName", Required = Required.DisallowNull)]`, which prevents you from serializing the object if `MiddleName` is `null`. But if `MiddleName` is `null`, I just want it to be absent from the JSON.

So the code as it stands doesn't really give you the ability to have optional attributes. Every property must always be present in the JSON (whether or not it's `null`).

rsuter:
Are we talking about plain JSON Schema or Swagger JSON Schema? Their handling is different...

rsuter:
... but yes it seems wrong...

tstrass:
I am just talking about plain JSON Schema. I am unfamiliar with Swagger JSON Schema.

rsuter:
Ok, we should change PropertyModel.JsonPropertyRequired: 

image

and PropertyModel.RenderRequiredAttribute

image

and the method: _property.IsNullable for the path "not Swagger handling".

rsuter:
Ok I think I've fixed this so that nothing breaks... I also improved the JSON Schema generator so that it handles DisallowNull... See https://github.com/NJsonSchema/NJsonSchema/commit/2e1af317a84e43ffdc980d08ecdbb6c511e91416

Also check out the unit test: 

https://github.com/NJsonSchema/NJsonSchema/blob/2e1af317a84e43ffdc980d08ecdbb6c511e91416/src/NJsonSchema.CodeGeneration.Tests/CSharp/CSharpGeneratorTests.cs#L683

Clone the project and compile the branch "TemplateFactory", then check if it is working for you...

rsuter:
V4.0 released, please try it and report here

tstrass:
Okay, looks good to me. :+1: 

And by the way NJsonSchema has been extremely helpful in my project. Thanks!

PR #174, opened 2016-08-16
Fix bug with ref that is file reference plus json path

tstrass:
This fixes a bug that causes an exception to be thrown when a `$ref` is a file reference plus a JSON path, e.g.
```JSON
{ "$ref": "path/to/file.json#/definitions/foo" }
```

Previously, the "#" was stripped out, so the JSON path part of the reference did not resolve. This change just leaves it in.

@rsuter I'm not sure of your process, but if let me know if you'd like me to create an issue for this bug and for #172


rsuter:
Thanks for the PR

PR #183, opened 2016-09-01
Added support for DataTypeAttribute

nickcox:


src/NJsonSchema/Generation/JsonSchemaGenerator.cs
```diff
@@ -531,6 +539,27 @@ public void ApplyPropertyAnnotations(JsonSchema4 jsonProperty, Type parentType,
                     jsonProperty.MaxLength = stringLengthAttribute.MaximumLength;
                 }
             }
+
+            dynamic dataTypeAttribute = TryGetAttribute(attributes,
+                "System.ComponentModel.DataAnnotations.DataTypeAttribute");
+            if (dataTypeAttribute != null)
+            {
+                var knownFormats = new Dictionary<string, string>
```
rsuter: You should make this static in the class so that it is not created multiple times...

rsuter:
Looks good so far, check my comment on line 547

rsuter:
You should also use the consts from here: 

https://github.com/NJsonSchema/NJsonSchema/blob/master/src/NJsonSchema/JsonFormatStrings.cs

Are your "format" strings official (ie. "data-url" and "phone")?

rsuter:
Also, we should either implement validation for these new format types or add an issue for this...

nickcox:
AFAICT the [official format strings](http://json-schema.org/latest/json-schema-validation.html#anchor107) are:

* date-time
* email
* hostname
* ipv4
* ipv6
* uri

However, custom formats are [allowed](http://json-schema.org/latest/json-schema-validation.html#anchor106) and these all seem to be supported by at least [some libraries](https://github.com/mozilla-services/react-jsonschema-form). I guess if we need to implement corresponding validations then that's an argument for not including the unofficial ones.

rsuter:
I think we should only validate the official ones and generate the formats defined by Json.NET

nickcox:
Makes sense.

rsuter:
Added new issue: https://github.com/NJsonSchema/NJsonSchema/issues/184

PR #193, opened 2016-09-15
Take into account the name of JsonPropertyAttribute when PropertyNameHandling is CamelCase

rogerbramon:
Currently, when PropertyNameHandling is CamelCase then the name of JsonPropertyAttribute is not used as json property name. In this PR we do the same than the Default case but applying camelcase after that.

### DIFF
--- a/src/NJsonSchema.Tests/Generation/PropertyNamesGenerationTests.cs
+++ b/src/NJsonSchema.Tests/Generation/PropertyNamesGenerationTests.cs
@@ -13,10 +13,10 @@ namespace NJsonSchema.Tests.Generation
         [DataContract]
         public class Foo
         {
-            [JsonProperty("BARbar1")]
+            [JsonProperty("BarBar1JsonProperty")]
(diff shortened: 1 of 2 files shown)


rsuter:
Are you sure that this is the way Json.NET behaves?

rogerbramon:
Yes, Json.NET work as is. I created an example:

Having this class,
```csharp
public class MyJsonPropertyTestClass
{
    [JsonProperty("NewName")]
    public string Name { get; set; }
}
```
and generating the schema as follows:

```csharp
var generator = new JSchemaGenerator()
{
    ContractResolver = new CamelCasePropertyNamesContractResolver()
};

var schema = generator.Generate(typeof(MyJsonPropertyTestClass));
```
it generates this schema:

```
{
  {
    "type": "object",
    "properties": {
      "newName": {
        "type": [
          "string",
          "null"
        ]
      }
    },
    "required": [
      "newName"
    ]
  }
}
```

rsuter:
Thank you for the PR

PR #268, opened 2016-12-10
Added line number information to errors

z16:
This PR adds support for line number/position in validation errors. This can, for example, be integrated in Visual Studio to give clickable errors in the output which then go right to the JSON file line and position, just like with C# errors.

To do this I had to change how the JToken is initialized. I no longer use `JsonConvert.Deserialize<JToken>`, but instead `JToken.Parse`. This required removing the `DateParseHandling = DateParseHandling.None`, since `JToken.Parse` uses `JsonLoadSettings` instead of `JsonSerializerSettings`. However, the default behavior of `JToken.Parse` uses the same date parsing handler, i.e. it just reads in the strings instead of doing any conversions.

If people do not want to parse line information (although the extra cost is negligible when compared to the validation logic) they have that option by calling `Validate` on a `JToken` they parsed themselves with their own settings.

As for the design, I added a `HasLineInfo` method to mirror `IJsonLineInfo`. The default values are `0`, again as with `IJsonLineInfo`, to make usage familiar to people who have worked with that part of Json.NET before.

I also added unit tests that cover all added features, although not in all possible combinations, only with common JSON validation errors, since the specific kind of error should not make a difference in how the data is parsed.

### DIFF
--- a/src/NJsonSchema.Tests/NJsonSchema.Tests.csproj
+++ b/src/NJsonSchema.Tests/NJsonSchema.Tests.csproj
@@ -74,6 +74,7 @@
     <Compile Include="Generation\TypeMappingsTests.cs" />
     <Compile Include="Serialization\ExtensionDataTests.cs" />
     <Compile Include="Validation\EnumValidationTests.cs" />
+    <Compile Include="Validation\LineInformationTest.cs" />
     <Compile Include="Validation\NullPropertyTests.cs" />
     <Compile Include="Validation\PatternPropertyValidationTests.cs" />
     <Compile Include="Validation\FormatBase64Tests.cs" />
--- a/src/NJsonSchema.Tests/Validation/LineInformationTest.cs
+++ b/src/NJsonSchema.Tests/Validation/LineInformationTest.cs
@@ -71,7 +71,7 @@ namespace NJsonSchema.Tests.Validation
         {
             foreach (var error in errors)
             {
-                Assert.AreEqual(hasLineInfo, error.HasLineInfo(), "HasLineInfo incorrect.");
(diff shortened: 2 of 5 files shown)


rsuter:
Thank you for the PR. It looks good. I think you should convert the HasLineNumber() method to a property and remove the _hasLineInfo member... Or is there a specific reason for this?

z16:
Only to keep it in line with [`IJsonLineInfo`](http://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_IJsonLineInfo.htm). I'm not a fan of it either and would rather it be a property, if you don't want it to resemble that interface, then I'm all for changing it.

rsuter:
I'd change it to a property... in the end, it is an own class and not having the exact same API is not a problem (in my opinion)

z16:
I agree :) Changed it now.

rsuter:
Do you need an immediate release of this PR?

z16:
No, I still have my local version I'm using, so I'm in no rush. Thanks for asking.

amoedoamorim:
I have tried it and it seems that the line info found on the error comes relative to the failed to validate object, rather than the error itself.

For instance: My object starts to be defined in line 1, which has an invalid additional property on line 5. When I try to get the LineNumber from the error, it says 1.

Is this the desired/intended behavior?

z16:
No, that's something I had not encountered before. I can take a look at it over the holidays, if no one does before, but I'll be busy for the rest of the week.

z16:
@amoedoamorim I adjusted it for the additional property validation. Did you notice it displaying that behavior on anything else?

PR #289, opened 2017-01-08
Introduce support for xmlObject

emilw:
Hi Rico,
As we discussed in the issue(https://github.com/NSwag/NSwag/issues/508) regarding support for xmlObject.
I have done the changes from the XMLSerializer:s point of view and tried to map it against the Swagger spec:
http://swagger.io/specification/#xmlObject

Can you have a look at my changes and give feedback?

I need to run some tests from NSwag to get the full file running, so i can really verify it in Swagger UI.
Or is there anything in NJsonSchema solution that can generate a full Swagger JSON?

This is my first pull request, so please let me know if i'm doing something wrong.

/Emil

Feedback (rsuter): 

- [x] **Can we create a second class with your new code and use the class in the JsonSchemaGenerator?**
-- Ok, i moved it to an extension, the same way as you have done with the XML docs. Please have a look what you think.
--- Yes, this is very good.
- [x] **I haven't tested the code, but I think the XML object is always created. Is this correct? Can we change the code, so that the XML object is only created if it really contains additional information?**
-- Response: Only for line items where it is needed to generate the <type></type> all the time, if the attributes are used to decorate the array property they are called afterward. Can you point in code, just so i do not misunderstand you?
--- It seems ok for me, will do some tests with the Swagger generator later...
- [x] **Rename GenerateXmlObject to GenerateXmlObjects**
-- Why do you want to have it in plural? It's only one xmlObject created per schema instance...
--- Because it can potentially generate multiple XML objects (for sub-/property schemas)
- [ ] **Why inherits JsonXmlObject from JsonSchema4?**
-- That is true, i reverse engineered your JsonProperty class and try to keep as close to that one as possible. This is changed now.
--- Why does it even need more properties than (name, namespace, prefix, attribute, wrapped) or a base class? I don't see why it has to inherit from JsonProperty or JsonSchema4...
- [ ] **ParentSchema should be set in the setter of the XmlObject property of JsonSchema4, this way it is always set even if a schema is manually created in code.**
-- I get a lot of tests failing when doing that, i need to look into why. I checked it in with it commented out.
--- Maybe you need a null check?
- [x] **Maybe the call to GenerateXmlObjectForItemTypeAsync must be before RequiresSchemaReference?**
-- I have moved it, I don't see the full impact of it though, have not full understanding of the impact.
--- Because a schema reference is needed (and if evaluates to true) then the XML object is not generated...
- [x] **Why are all the methods async? Creating a new XML object should not be created in a task (it is CPU-bound). I think we should remove the async part of the methods - except if we have I/O somewhere in the code (i.e. loading a remote xml file?)?**
-- True, I have changed that.
--- Thanks

### DIFF
--- a/src/NJsonSchema/JsonSchema4.cs
+++ b/src/NJsonSchema/JsonSchema4.cs
@@ -458,7 +458,20 @@ namespace NJsonSchema
 
         /// <summary>Gets the xml object of the type. </summary>
         [JsonProperty("xml", DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
-        public JsonXmlObject Xml { get; set; }
+        public JsonXmlObject Xml {
+            get
+            {
+                return _xmlObject;
+            }
+            set
+            {
+                _xmlObject = value;
+                //_xmlObject.ParentSchema = this;
+            }
+        }
+
+        [JsonIgnore]
+        private JsonXmlObject _xmlObject;
 
         /// <summary>Gets the pattern properties of the type. </summary>
         [JsonIgnore]
--- a/src/NJsonSchema/JsonSchema4.cs
+++ b/src/NJsonSchema/JsonSchema4.cs
@@ -456,6 +456,10 @@ namespace NJsonSchema
             }
         }
 
+        /// <summary>Gets the xml object of the type. </summary>
+        [JsonProperty("xml", DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        public JsonXmlObject Xml { get; set; }
+
         /// <summary>Gets the pattern properties of the type. </summary>
         [JsonIgnore]
         public IDictionary<string, JsonSchema4> PatternProperties
--- a/src/NJsonSchema.Tests/Generation/XmlObjectTests.cs
+++ b/src/NJsonSchema.Tests/Generation/XmlObjectTests.cs
@@ -1,0 +1,250 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NJsonSchema.Infrastructure;
+using Newtonsoft.Json.Linq;
+using System.Xml.Serialization;
+
+namespace NJsonSchema.Tests.Generation
+{
+    [TestClass]
+    public class XmlObjectTests
+    {
+        private const string StringArray = "StringArray";
+        private const string IntArray = "IntArray";
+        private const string DoubleArray = "DoubleArray";
+        private const string DecimalArray = "DecimalArray";
+        private const string InternalItemArray = "InternalItem";
+        private const string Foo = "Foo";
+
+        public class WithoutXmlAttributesDefined
+        {
+            public string Foo { get; set; }
+            public string[] StringArray { get; set; }
+            public int[] IntArray { get; set; }
+            public double[] DoubleArray { get; set; }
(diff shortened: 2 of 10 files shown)


rsuter:
Hi Emil, 

Thank you for the PR, it looks very good!

I have the following points: 

- Can we create a second class with your new code and use the class in the serializer?
- Why are all the methods async? Creating a new XML object should not be created in a task (it is CPU-bound). I think we should remove the async part of the methods - except if we have I/O somewhere in the code (i.e. loading a remote xml file?)?
- I haven't tested the code, but I think the XML object is always created. Is this correct? Can we change the code, so that the XML object is only created if it really contains additional information?

My development workflow: 

1. Create a new directory abc
2. Clone NJsonSchema repo into abc
3. Clone NSwag repo into abc
4. Use NuGetReferenceSwitcher to switch the NuGet references in NSwag to local projects: https://github.com/rsuter/NuGetReferenceSwitcher

Changes in NSwag: 

To test your code, just change the settings class in NSwag and set the GenerateXmlObject to true in the ctor (like NullHandling):

image

src/NJsonSchema/JsonSchema4.cs
```diff
@@ -456,6 +456,23 @@ internal set
...
+        [JsonProperty("xml", DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
+        public JsonXmlObject Xml {
+            get
+            {
+                return _xmlObject;
+            }
+            set
+            {
+                _xmlObject = value;
+                //_xmlObject.ParentSchema = this;
```
emilw: When enabling this line, i get the failing tests. I will check into it later.

emilw:
Hi,
Thanks for feedback!
I have responded in my original comment, please have a look so i understood you correctly.
I still have not had time to run you development workflow, i will try to do that when i get time.
/Emil

rsuter:
Thanks for working on NJS, ... I hope you are not tired yet :-). I have updated the comments... looks very good

rsuter:
What is the purpose of 

    internal static JsonXmlObject FromJsonSchema(string name, JsonSchema4 type)

rsuter:
Maybe we should rename JsonXmlObject to SwaggerXmlObject?

src/NJsonSchema/JsonSchema4.cs
emilw: It was the null check...

emilw:
Hi,
I have updated my initial comment with answers. I think all bullets are closed now.

- The FromJsonSchema was a left over from deriving from JsonSchema4, it's removed.
- I changed the name, now when i understand that it is not a part of JsonSchema4 spec

I also got a positive response here around the wrapped property:
https://github.com/swagger-api/swagger-ui/issues/2601

So the code is right from that point of view also.

I will see if i manage to run the tests with NSwag this week.

/Emil

rsuter:
I already fixed everything and merged it into NJS and updated NSwag... you can try NSwag with the latest Ci build artifacts...

emilw:
Aha, great, I did not get that :-).
I checked your stuff and the null check is not in your master and the renaming to SwaggerXmlObject, i have pushed my last commit if you want to compare with yours.

rsuter:
The null check is in master and I think we'll leave JsonXmlObject as is (or better: I have to think about this :-)) But its working now...

emilw:
Ok, yes I see it now...i checked the merge commit :-(.

Issue #291, opened 2017-01-09
Schema generation fails when a KnownType attribute specifies a method name instead of a type


csdahlberg:
If a type uses a `KnownType` attribute to specify the name of a method that returns known types, a `NullReferenceException` is thrown when `JsonSchemaGenerator.GenerateKnownTypesAsync` calls `JsonObjectTypeDescription.FromType`. A simple example:

```
    public class Pencil
    {
    }

    [KnownType("GetKnownTypes")]
    public class WritingInstrument
    {
        public static Type[] GetKnownTypes()
        {
            return new[] { typeof(Pencil) };
        }
    }

    public class Container
    {
        public WritingInstrument WritingInstrument { get; set; }
    }

    public async Task ReproAsync()
    {
        var schema = await JsonSchema4.FromTypeAsync<Container>();
    }
```

I have support for this scenario coded, and plan on submitting a pull request shortly.

rsuter:
I didnt know that this is possible. Thanks for the PR.

PR #292, opened 2017-01-09
Added support for KnownType attributes that specify the name of a method

csdahlberg:
This is for issue #291.


rsuter:
Thanks for the PR

Issue #308, opened 2017-01-27
Retain data annotations on generated models


cbordeman:
Attributes such as [Required], [Range], [StringLength(5)] from the System.ComponentModel.DataAnnotations namespace should be retained in the generated client side models so validation can be performed.

rsuter:
https://github.com/NSwag/NSwag/commit/85a802083fad8833f5daf9e0a2d1626307c19633

RSuter:
I think C# only as TypeScript/JS has no common way to express this...

cbordeman:
Thanks very much for implementing this!

PR #318, opened 2017-02-15
Pull request for issue #308

Zehelein:
Hi,

I created the pull request for issue #308 to retain data annotations. I added logic for range, string length and regular expression data annotations. I did not implement the min/max items and unique items as they would require custom validators. Not sure how to implement that in this NJsonSchema context - maybe that could/should go to NSwag, maybe in NJsonSchema some custom attribute could be added but commented out?

Best regards,

Frank

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/CSharpGeneratorSettings.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp/CSharpGeneratorSettings.cs
@@ -22,8 +22,8 @@ namespace NJsonSchema.CodeGeneration.CSharp
             ArrayType = "System.Collections.ObjectModel.ObservableCollection";
             DictionaryType = "System.Collections.Generic.Dictionary";
 
-            RequiredPropertyMustBeDefined = true;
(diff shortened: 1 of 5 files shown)


rsuter:
Thank you for the PR, looks very good. 

I think we should rename "DataAnnotationsMustBeDefined" to "GenerateDataAnnotations"
And please don't rename "RequiredPropertiesMustBeDefined" as this is a breaking change

Zehelein:
Yes true - did not think about the breaking change. The "GenerateDataAnnotations" one is definitely fine. 

Should I do the changes and create another pull request or will you do that when you merge it?

rsuter:
You can do more commits in this PR (good to know: A PR only states which to branches to merge, the time when the PR is created does not matter, all commits are merged when the PR is accepted)

rsuter:
Otherwise I'll do it after the merge

Zehelein:
Ok - thanks for the information. I updated it now with those two changes.

cbordeman:
Thanks very much guys.  This is a nice feature!

PR #327, opened 2017-02-27
Fix for 324 - Exception when generating schema on types that inherit a known type

markledwich2:
When generating additional known types, the code was intending to look at classes that have the KnownType attribute set explicitly (not inherited).
 - Made fix to not look into inherited classes for the known type static method. 
- Updated tests to expose this problem/fix
- Avoided some exceptions that were getting swallowed finding xml documentation. It was making it more difficult to debug and was a minor change with no behavioral side effects i could think of.


rsuter:
Thank you, also for the other improvements...

PR #329, opened 2017-02-28
Fix for inheritance of abstract properties and ContractResolver that suppresses properties

markledwich2:
- fixed logic for which properties should be included in schema
- added test for custom contract resolver
- fixed issue with inheritance with an abstract property
- suppress warnings for xml comments in test project


markledwich2:
There is plenty of work to go using the parts from the Json.NET contract resolver that overlap with TypeDescription and other classes. I did the minimal I needed to for my project and I'm not confident in #d7a2429, but I think the unit test should be included as serializing an object to a string using the TypeConverter method is standard for Json.NET.

PR #345, opened 2017-03-23
Use custom contract resolver overrides for Dictionary and Array contracts 

markledwich2:
Changes
- added test for custom contract resolver that overrides default array contract
- only treat as array or dictionary when current NJsonSchema logic and contract resolver agree

The serialization by json.net in [regards to dictionaries and arrays](http://www.newtonsoft.com/json/help/html/SerializationGuide.htm#Breakdown) is not always desired. There are some cases that NJsonSchema differs (e.g. when you have extended a dictionary and added a property). When using both json.net and NJSonSchema and customizing serialization behavior, the contract resolver is a way to that both serialization and schema generation can be customized to do something differently.

These changes works towards customization being treated the same way for serialization and schema generation. Further changes should be made so that the custom treating of objects (e.g. dictionaries) by schema and code generation is done via custom contract revolvers rather than separate logic.

### DIFF
--- a/src/NJsonSchema.Tests/Generation/ContractResolverTests.cs
+++ b/src/NJsonSchema.Tests/Generation/ContractResolverTests.cs
@@ -57,9 +57,9 @@ namespace NJsonSchema.Tests.Generation
             protected override JsonContract CreateContract(Type objectType)
             {
                 JsonContract contract = base.CreateContract(objectType);
-
                 // by default a type that can convert to string and that is also an enum will have an array contract, but serialize to a string!. fix  this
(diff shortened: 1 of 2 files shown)


rsuter:
I think there's a compile error: 

image

https://ci.appveyor.com/project/rsuter/njsonschema-jlw0p/build/1135

markledwich2:
Apologies. I had some c#7 in there.

rsuter:
No problem. Thanks for the fix... Ill have to do more testing until we can merge the branch... I also have to change NSwag to support these changes.

Issue #294, opened 2017-01-10
Improve JsonInheritanceConverter 


rsuter:
The problem is, that the JsonInheritanceConverter does not handle the schema name correctly. The bugfix is to change the IsKnownTypeTargetType to:

	private bool IsKnwonTypeTargetType(dynamic attribute, string discriminator)
	{
		Type type = attribute.Type;

		if (type == null)
		{
			return false;
		}

		var typeName =
			type.GetTypeInfo().GetCustomAttribute<JsonSchemaAttribute>()?.Name ?? type.Name;

		return typeName == discriminator;
	}

But besides that I would prefer that the schema generation is indepenent of how I serialize and deserialize the json, e.g. with another (optional?) Attribute: [JsonDiscriminator]

See https://github.com/NSwag/NSwag/issues/455

SebastianStehle:
What are your plans for this? I would like to have a service and a default implementation that 
...just checks the KnownTypeAttribute and use
...uses type name
...uses $type as discriminator
...also supports the JsonInheritanceConverter 

=> compatible with JSON.NET default behavior.

:)

RicoSuter:
Is this still an issue?

PR #347, opened 2017-03-23
Add KnownType and JsonSchema annotations to generated C# models

dwickern:
Currently if you convert this class into schema...
```
[JsonConverter(typeof(JsonInheritanceConverter), "animalType")]
[KnownType(typeof(Dog))]
class Animal {}

[JsonSchema("custom-dog")]
class Dog : Animal {}
```

...you will get something like:
```
definitions:
  animal:
    type: object
    discriminator: animalType
  custom-dog:
    type: object
    allOf:
    - $ref: '#/definitions/animal'
```

But the conversion back to C# is lossy:
```
[JsonConverter(typeof(JsonInheritanceConverter), "animalType")]
class Animal {}
class CustomDog : Animal {}
```

JsonInheritanceConverter doesn't have enough information to deserialize `{ "animalType": "custom-dog" }`.

This PR adds `JsonSchema` and `KnownType` attributes to the generated C# class.

Related to #294 and NSwag/NSwag#455

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/Templates/ClassTemplate.tt
+++ b/src/NJsonSchema.CodeGeneration.CSharp/Templates/ClassTemplate.tt
@@ -5,6 +5,9 @@
 <#if(Model.HasDiscriminator){#>
 [Newtonsoft.Json.JsonConverter(typeof(JsonInheritanceConverter), "<#=Model.Discriminator#>")]
 <#}#>
+<#if(Model.HasCustomName){#>
+[NJsonSchema.Annotations.JsonSchema("<#=Model.CustomName#>")]
+<#}#>
 [System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "<#=JsonSchema4.ToolchainVersion#>")]
 public partial class <#=Model.Class#> <#=Model.Inheritance#>
 {
--- a/src/NJsonSchema.CodeGeneration.CSharp/Templates/ClassTemplate.tt
+++ b/src/NJsonSchema.CodeGeneration.CSharp/Templates/ClassTemplate.tt
@@ -8,6 +8,9 @@
 <#if(Model.HasCustomName){#>
 [NJsonSchema.Annotations.JsonSchema("<#=Model.CustomName#>")]
 <#}#>
+<#foreach(var knownType in Model.KnownTypes){#>
+[System.Runtime.Serialization.KnownType(typeof(<#=knownType#>))]
+<#}#>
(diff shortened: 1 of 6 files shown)


src/NJsonSchema.CodeGeneration.CSharp/Templates/ClassTemplate.tt
```diff
@@ -5,6 +5,12 @@
 <#if(Model.HasDiscriminator){#>
 [Newtonsoft.Json.JsonConverter(typeof(JsonInheritanceConverter), "<#=Model.Discriminator#>")]
 <#}#>
+<#if(Model.HasCustomName){#>
+[NJsonSchema.Annotations.JsonSchema("<#=Model.CustomName#>")]
```
rsuter: We shouldn't add NJsonSchema dependent attributes here... The code should work out-of-the-box.

dwickern: It looks like `[Newtonsoft.Json.JsonObject(Title = "...")]` is what we need.

rsuter:
And btw: JsonInheritanceConverter does not need knowntypeattrs because it also tries to find the type in the same namespace...

dwickern:
You're referring to [this code](https://github.com/NJsonSchema/NJsonSchema/blob/c43dca9f109bd0e9afb1a843f8256c0a678cac81/src/NJsonSchema.CodeGeneration.CSharp/Templates/JsonInheritanceConverterTemplate.tt#L99-L100)? If we try to deserialize `{ "animalType": "custom-dog" }`, the type `MyNamespace.custom-dog` doesn't exist. That's what I'm trying to fix with this PR.

rsuter:
Ok, yes. If it is renamed, this code will not work... Is 

    [Newtonsoft.Json.JsonObject(Title = "...")]

already supported by NJS? Or only the JsonSchemaAttribute?

I think we should add support for JsonObjectAtrribute.Title and generate that and KnownTypeAttribute

dwickern:
NJS only reads `JsonSchemaAttribute` so we'll need to add `JsonObjectAttribute.Title` support on both ends.

rsuter:
I think your pr does not really work. Also we need to apply this to typescript. 

First change: https://github.com/NJsonSchema/NJsonSchema/commits/master

dwickern:
I did some more research and think I better understand what we need. It ended up being more work than I expected. I opened #369 with a TODO list. I'll chip away as time permits

PR #348, opened 2017-03-26
ValueSource introdeced in the JsonSchemaExtensionDataAttribute

dmitriyse:
This improvement allows to put into the schema arbitrary extension data, not only build-in clr types.
Very useful to put rich "example" objects.

### DIFF
--- a/src/NJsonSchema/Annotations/JsonSchemaPostProcessAttribute.cs
+++ b/src/NJsonSchema/Annotations/JsonSchemaPostProcessAttribute.cs
@@ -1,5 +1,5 @@
-﻿//-----------------------------------------------------------------------
-// <copyright file="JsonSchemaAttribute.cs" company="NJsonSchema">
+//-----------------------------------------------------------------------
+// <copyright file="JsonSchemaPostProcessAttribute.cs" company="NJsonSchema">
 //     Copyright (c) Rico Suter. All rights reserved.
 // </copyright>0
 // <license>https://github.com/rsuter/NJsonSchema/blob/master/LICENSE.md</license>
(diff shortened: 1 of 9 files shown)


src/NJsonSchema/Annotations/JsonSchemaPostProcessAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+// <author>Rico Suter, mail@rsuter.com</author>
+//-----------------------------------------------------------------------
+namespace NJsonSchema.Annotations
+{
+    using System;
+
+    /// <summary>
+    /// Allows to post process schema.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Property, AllowMultiple = true)]
```
RSuter: Attribute usage should only be class, otherwise multiple processors per type could be registered which may lead to strange behaviors

src/NJsonSchema/Generation/JsonSchemaGenerator.cs
```diff
@@ -195,19 +230,50 @@ public Task<TSchemaType> GenerateAsync<TSchemaType>(Type type, JsonSchemaResolve
...
+            {
+                schema.ExtensionData = new Dictionary<string, object>();
+                foreach (var attribute in extensionDataAttributes)
+                {
+                    TryAddExtensionData(schema.ExtensionData, attribute);
+                }
+            }
+        }
+
+        private void TryAddExtensionData(IDictionary<string, object> extensionData, JsonSchemaExtensionDataAttribute attribute)
```
RSuter: Move method into JsonSchemaExtensionDataAttribute class?

RSuter:
Can't you just implement a new attribute which inherits JsonSchemaExtensionDataAttribute and add your custom logic to it's processor?

PR #376, opened 2017-05-03
Added Schema property to ValidationError

sunahe:
Added Schema property to ValidationError so that it is easier to find the constraint that caused the validation error.

This will allow better localization of the validation errors.
For instance, one can extend the json schema:
```
{
  "properties": {
    "name": {
      "type": "string",
      "regex": "^[\\\\w\\\\d]*$",
      "x-regexMessage": "The name can only contain alphanumeric characters"
    }
  }
}
```

... and then use the extensions from validation errors:

```
ValidationError error = schema.Validate(obj).FirstOrDefault();
if (error?.Kind == ValidationErrorKind.PatternMismatch)
  errorMessage = error.Schema.ExtensionData["x-regexMessage"].ToString();
```


rsuter:
Thanks for the PR. I think this is a good addition to the validator...

PR #399, opened 2017-06-08
Object is newable only if it's not an enumeration

exyi:
NSwag generates `<enum name>.fromJS(...)` for string-based enum types that do not have the `fromJS` method, so this should fix it.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.TypeScript/DataConversionGenerator.cs
+++ b/src/NJsonSchema.CodeGeneration.TypeScript/DataConversionGenerator.cs
@@ -107,7 +107,7 @@ namespace NJsonSchema.CodeGeneration.TypeScript
         {
             schema = schema.ActualSchema;
             return (schema.Type.HasFlag(JsonObjectType.Object) || schema.Type == JsonObjectType.None) 
-                && !schema.IsAnyType && !schema.IsDictionary;
+                && !schema.IsAnyType && !schema.IsDictionary && !schema.IsEnumeration;
         }
     }
 }


RSuter:
Is this a fix for https://github.com/NSwag/NSwag/issues/752 ?

exyi:
Seems like it is the same issue. for example, this swagger was broken: https://api.apis.guru/v2/specs/github.com/v3/swagger.json

Thank you merge

RSuter:
Updating NSwag now...

RSuter:
Thanks for the pr... Strange that i didnt see this bug in my projects...

exyi:
I think it bugs only if the enum is string-based, int-based are *probably* OK. But I have not tested...

RSuter:
Is this spec now fully working?

RSuter:
https://github.com/NSwag/NSwag/commit/3e4f2216a63cb1ade27dd4818759afc4c8e7bffc

exyi:
Unfortunately ;( By changing the Type of the object to String in SwaggerDocument I could get it working on some places (I don't know why...), but this fix actually did not change anything in the document. I'm sorry for that.

RSuter:
So this problem is still not fixed?

RSuter:
Now this compiles for me: 

- https://github.com/NSwag/NSwag/blob/master/src/NSwag.Sample.NetCoreAngular/github-nswag.nswag
- https://github.com/NSwag/NSwag/blob/master/src/NSwag.Sample.NetCoreAngular/ClientApp/app/githubServices.ts
- https://github.com/NSwag/NSwag/blob/master/src/NSwag.Sample.NetCoreAngular.Clients/GitHubClients.cs

exyi:
This looks good, I probably did the update badly, sorry. Thank you for your effort, I'll try it later.

PR #409, opened 2017-06-21
Allow type discovery to fallback to $type

paulomorgado:


RSuter:
Sorry, I had to revert some changes because they were not supported in .net 4.0... Additionally I had to expand all types to fulltype names (no usings allowed) and copied over the code into the template...

RSuter:
Sorry, I have to remove

```
        var typeInfo = jObject.GetValue("$type");
        if (typeInfo != null)
            return System.Type.GetType(Newtonsoft.Json.Linq.Extensions.Value<string>(typeInfo));
```

because it is a security hole...

RSuter:
Sorry I had to revert the changes because it broke the CI build and I actually don't see why this is needed - if the JSON respects the spec, then the discriminator is available or a KnownTypeAttribute set...

paulomorgado:
The goal is to handle "unknown types" that derive from the known type.

What security problems are you talking about?

PR #413, opened 2017-06-24
Introduce SIngleton type

Bert-Proesmans:
I'm currently working with JSON schemas and encountered a use case where an object property is allowed to contain ONLY ONE of the provided objects. Which I named the Singleton type, but another name might fit better.

The generated code declares the property type of a Singleton as a (generic) object, which is enough since resolving/downcasting that object falls under application logic.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/CSharpTypeResolver.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp/CSharpTypeResolver.cs
@@ -39,7 +39,7 @@ namespace NJsonSchema.CodeGeneration.CSharp
         {
             schema = schema.ActualSchema;
 
-            if (schema.IsAnyType)
+            if (schema.IsAnyType || schema.IsSingleton)
                 return "object";
 
             var type = schema.Type;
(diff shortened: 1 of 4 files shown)


RSuter:
Hi, thanks for the PR. But I don't think using "object" or "any" if a schema is a "singleton" is correct.

https://github.com/RSuter/NJsonSchema/issues/13

In your sample, it is only any, if the oneOf schema is also any... Maybe we just have to check in IsAnyType: 

   IsAnyType => [...current logic] && OneOf.All(s => s.IsAnyType)

Bert-Proesmans:
You're right, a lot of information is lost by using object. I thought about it for some time but couldn't find a clean way to implement it in C#. I haven't used typescript so i can't attribute to the topic regarding it.

As far as i know the options are building a C-like enum [[1]], using templates [[2]], or a very verbose wrapper class [[1]-Peter N Roth's answer]. All three options have downsides and a compromis should be made for the least unnatural way of writing C#.

My preference lies not with the template option.

[1]:https://social.msdn.microsoft.com/Forums/en-US/60150e7b-665a-49a2-8e2e-2097986142f3/c-equivalent-to-c-union?forum=csharplanguage
[2]:https://stackoverflow.com/a/7302957

RSuter:
Your unit test passes even when no other changes are made in the codebase... What exactly is your problem? 

I think oneOf should be handled with inheritance, i.e. all classes/schemas in oneOf inherit from the base schema...

PR #415, opened 2017-06-26
All numeric types supported in schema default values conversions to C# and TypeScript

tomasherceg:
I have added support for all numeric types in the `DefaultValueGenerator` for C# and TypeScript with tests, as we discussed in the [NSwag PR 818](https://github.com/NSwag/NSwag/pull/818).


RSuter:
Just one question: What is the type of schema.Default? Isnt it a JToken instead of a primitive value sometimes?

RSuter:
Problem: https://github.com/RSuter/NJsonSchema/pull/458

PR #437, opened 2017-07-21
time format fixes RSuter/NJsonSchema#436

### LINKED ISSUE #436: Time format is not correct according to RFC 3339
I see that the time format has been added in #423. The format for the time is not exactly correct. It does not follow either the old v3 json schema time format or the RFC 3339 time format. It should allow either of these.

According the RFC 3339, a time format should be:

```
   time-hour       = 2DIGIT  ; 00-23
   time-minute     = 2DIGIT  ; 00-59
   time-second     = 2DIGIT  ; 00-58, 00-59, 00-60 based on leap second rules
   time-secfrac    = "." 1*DIGIT
   time-numoffset  = ("+" / "-") time-hour ":" time-minute
   time-offset     = "Z" / time-numoffset

   partial-time    = time-hour ":" time-minute ":" time-second [time-secfrac]
   full-time       = partial-time time-offset
```

examples include 

- 14:30:00+02:00
- 14:30:00-02:00
- 14:30:00Z

The old v3 json schema draft simply said "time-This SHOULD be a time in the format of hh:mm:ss"
- 14:30:00

None of these formats are supported by the currently used format string which is 
```HH:mm:ss.FFFFFFF zzz``` 
it also **requires** an inappropriate space and a timezone offset.

Switching to the format string 
```"HH:mm:ss.FFFFFFFK"```
the K modifier will allow for either 

- utc time zone (Z)
- a time zone offset such as +05:00
- an empty string when no time zone is specified.

https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings#KSpecifier

I can create a pull request with the change and tests for the various time formats

DanSmith:
This pr validates the time format as specified in json schema v3 and in RFC 3339

https://tools.ietf.org/html/draft-zyp-json-schema-03
https://tools.ietf.org/html/rfc3339#section-5.6

This fixes #436

### DIFF
--- a/src/NJsonSchema.Tests/Validation/FormatTimeTests.cs
+++ b/src/NJsonSchema.Tests/Validation/FormatTimeTests.cs
@@ -5,6 +5,10 @@ using NJsonSchema.Validation;
 
 namespace NJsonSchema.Tests.Validation
 {
+    /// <summary>
+    /// Time format tests according to rfc 3339 full-time
+    /// https://tools.ietf.org/html/rfc3339#section-5.6
+    /// </summary>
     [TestClass]
     public class FormatTimeTests
     {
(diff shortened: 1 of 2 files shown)


cortex93:
looks better. don't know why i add extra space by reading the rfc

PR #448, opened 2017-08-17
Fix for stack overflow on Linux.

### LINKED ISSUE #447: JsonSchema4.FromJsonAsync() can stack overflow on Linux
The `JsonSchema4.FromJsonAsync()` function can stack overflow on Linux when presented with the following schema:

```json
{
  "x-datetime": "2016-07-28T14:39:37.937Z"
}
```

In `JsonSchemaReferenceUtilities.UpdateSchemaReferencesAsync()` it appears to recursively enumerate all properties including those which are `static`.

As the above string is interpreted as a `DateTime`, this means that `DateTime.Now` is visited, and despite being added to the cache, because it is liable to change on every invocation, the check for already visited objects fails to prevent recursion which ends up causing a stack overflow.

I've been unable to reproduce the issue on Windows unless I artificially introduce a delay to cause `DateTime.Now` to return different values for each invocation.

Sumo-MBryant:
The `DateTime.Now` static property was being visited
which may return a unique value every time, allowing
unbounded recursion.

Fixes #447.

### DIFF
--- a/src/NJsonSchema.Tests/Infrastructure/ReflectionCacheTests.cs
+++ b/src/NJsonSchema.Tests/Infrastructure/ReflectionCacheTests.cs
@@ -1,0 +1,34 @@
+using System.Linq;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NJsonSchema.Infrastructure;
+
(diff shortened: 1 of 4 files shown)


RSuter:
Thank you for the pr.

PR #458, opened 2017-09-06
Test that shows CSharpGenerator currently fails for integer properties with default values

bobend:
Tries to assign a long to an int property.
Currently generates this invalid c# code:
```csharp
public int IntergerWithDefault { get; set; } = 5L;
```


RSuter:
Tasks: 

- [ ] Add schema parameter to ConvertNumericValue() and check ActualSchema.Type and ActualSchema.Format instead of object type, https://github.com/RSuter/NJsonSchema/blob/dcd3cc8189e082ef2457d2627e22af48c04ffcb0/src/NJsonSchema.CodeGeneration/DefaultValueGenerator.cs#L58

RSuter:
Ref https://github.com/RSuter/NJsonSchema/pull/415

RSuter:
Thank you for the PR.

PR #493, opened 2017-10-02
Implement 'date' type handling for Typescript CodeGeneration

### LINKED ISSUE #492: type: date is not formatted correctly in typescript
Similar to #491, but when generating a typescript class with moment field types (or Date fields).

When converting a `"type": "date"` field to json, we call `DateToStringCode -> toISOString()`, which generates a date-time, not a date.

ref: https://swagger.io/specification/#document-structure-12

danzel:
Fixes #492 

We now have `IsDate` and `IsDateTime`, and use these to generate different code when serializing back to json.

`DateTimeCodeGenerationTests.cs` is the previous `DateCodeGenerationTests.cs` and matches it other than by name.
`DateCodeGenerationTests.cs` is a new lot of tests that check the same things, but for the `date` type.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.Tests/TypeScript/DateCodeGenerationTests.cs
+++ b/src/NJsonSchema.CodeGeneration.Tests/TypeScript/DateCodeGenerationTests.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+﻿using System.Threading.Tasks;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 using NJsonSchema.CodeGeneration.TypeScript;
 
 namespace NJsonSchema.CodeGeneration.Tests.TypeScript
@@ -8,17 +9,18 @@ namespace NJsonSchema.CodeGeneration.Tests.TypeScript
     {
         private const string Json =
 @"{
-	""type"": ""object"", 
-	""properties"": {
-		""myDate"": ""2017-01-01""
+    '$schema': 'http://json-schema.org/draft-04/schema#',
+	'type': 'object', 
+	'properties': {
+		'myDate': { 'type': 'string', 'format': 'date' }
 	}
 }";
 
         [TestMethod]
-        public void When_date_handling_is_string_then_string_property_are_generated_in_class()
+        public async Task When_date_handling_is_string_then_string_property_are_generated_in_class()
         {
             //// Arrange
-            var schema = JsonSchema4.FromData(Json);
+            var schema = await JsonSchema4.FromJsonAsync(Json);
 
             //// Act
             var generator = new TypeScriptGenerator(schema, new TypeScriptGeneratorSettings
@@ -35,10 +37,10 @@ namespace NJsonSchema.CodeGeneration.Tests.TypeScript
         }
 
         [TestMethod]
-        public void When_date_handling_is_moment_then_moment_property_are_generated_in_class()
+        public async Task When_date_handling_is_moment_then_moment_property_are_generated_in_class()
         {
             //// Arrange
-            var schema = JsonSchema4.FromData(Json);
+            var schema = await JsonSchema4.FromJsonAsync(Json);
 
             //// Act
             var generator = new TypeScriptGenerator(schema, new TypeScriptGeneratorSettings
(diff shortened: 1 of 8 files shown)


src/NJsonSchema.CodeGeneration.Tests/TypeScript/DateCodeGenerationTests.cs
```diff
@@ -1,23 +1,24 @@
...
+	}
+}";
 
         [TestMethod]
-        public async Task When_date_handling_is_string_then_string_property_are_generated_in_class()
+        public void When_date_handling_is_string_then_string_property_are_generated_in_class()
         {
             //// Arrange
-            var schema = await JsonSchema4.FromTypeAsync<ClassWithDateProperty>();
+            var schema = JsonSchema4.FromData(Json);
```
RSuter: FromData() is wrong here - FromData() creates a schema from sample data, in this case this would be a schema with the properties "type" and "properties". Either we use "FromJson()" here or we use "FromTypeAsync" (like in the other tests) and add the [JsonSchema("string", Format = "Date")]` attribute

RSuter:
Thank you for the PR - this looks like one of the high quality PR (with unit tests, no breaking changes, fixed for all paths [date, moment, string], etc.). Please see my review comment - this will probably fix the unit tests. Then we can merge...

src/NJsonSchema.CodeGeneration.TypeScript/DataConversionGenerator.cs
```diff
@@ -72,22 +74,21 @@ private static object CreateModel(DataConversionParameters parameters)
                 ArrayItemType = parameters.Resolver.TryResolve(parameters.Schema.Item, parameters.TypeNameHint) ?? "any",
                 IsArrayItemNewableObject = parameters.Schema.Item != null && IsNewableObject(parameters.Schema.Item),
                 IsArrayItemDate = IsDate(parameters.Schema.Item?.Format, parameters.Settings.DateTimeType),
+                IsArrayItemDateTime = IsDateTime(parameters.Schema.Item?.Format, parameters.Settings.DateTimeType),
 
                 StringToDateCode = parameters.Settings.DateTimeType == TypeScriptDateTimeType.Date ? "new Date" : "moment",
-                DateToStringCode = "toISOString()", 
+                DateToStringCode = "toISOString().slice(0, 10)",
```
RSuter: Is this safe? Looks a little ugly to me :-)

src/NJsonSchema.CodeGeneration.TypeScript/Templates/ConvertToClassTemplate.tt
```diff
@@ -9,7 +9,7 @@ if (<#=Model.Value#> && <#=Model.Value#>.constructor === Array) {
 <#if(Model.IsArrayItemNewableObject){#>
         <#=Model.Variable#>.push(<#=Model.ArrayItemType#>.fromJS(item<#if(Model.HandleReferences){#>, _mappings<#}#>));
 <#}else{#>
-<#if(Model.IsArrayItemDate){#>
+<#if(Model.IsArrayItemDate || Model.IsArrayItemDateTime){#>
         <#=Model.Variable#>.push(<#=Model.StringToDateCode#>(item));
```
RSuter: Maybe we should also add `StringToDateTimeCode` (like in the other direction) - it may be more (redundant) code, but it may be clearer for the next who looks into this code?

src/NJsonSchema.CodeGeneration.TypeScript/DataConversionGenerator.cs
danzel: Haha, yeah this works. It's a bit ugly, but it works for `Date` and `moment` types:
```js
moment().toISOString().slice(0, 10)
"2017-10-02"
new Date().toISOString().slice(0, 10)
"2017-10-02"
```
Luckily `toISOString` is well defined :)

src/NJsonSchema.CodeGeneration.TypeScript/Templates/ConvertToClassTemplate.tt
danzel: Or I could put a comment on `StringToDateCode` that it is used for both Date and DateTime conversion?
I don't think there would be a reason that these would need to be different.
I'm happy either way, let me know.

src/NJsonSchema.CodeGeneration.Tests/TypeScript/DateCodeGenerationTests.cs
danzel: I swear this passed locally!
Will change as suggested tomorrow.

danzel:
I think the tests failed because I hadn't committed the updated .cs files generated from the .tt files.
Have made all of the requested changes and the tests pass now :+1:

RSuter:
Thanks for the pr. Keep em coming 👍

PR #496, opened 2017-10-03
Support 'date' format properties in the CSharp Generator

### LINKED ISSUE #491: type: date format is not formatted correctly in c#
When a `"type": "date"` property is present in a schema we can generate a DateTime or DateTimeOffset field.
However, when when serializing this field it is not turned in to a date, it is turned in to a dateTime.

ref: https://swagger.io/specification/#document-structure-12
date should be a full-date, which is `yyyy-MM-dd` in c#.

To fix this, you can add `[Newtonsoft.Json.JsonConverter(typeof(DateFormatConverter))]` to the Property and define that converter as:
```
	public class DateFormatConverter : Newtonsoft.Json.Converters.IsoDateTimeConverter
	{
		public DateFormatConverter()
		{
			DateTimeFormat = "yyyy-MM-dd";
		}
	}
```

Trying to figure out how to fix this myself, but... there is a lot of code in here :)

danzel:
Fixes #491 

Like #493, but for csharp :)


RSuter:
Thank you - looks good! Just made some minor adjustments: 

- Made DateFormatConverterTemplateModel public: Will be public in API when we have a new template engine, what do you think? => https://github.com/RSuter/NSwag/issues/263
- Use JsonFormatStrings.Date instead of "date"

PR #497, opened 2017-10-03
Serializer setting

lisemon:
Hmm, how about something like this? 😄

### DIFF
--- a/src/NJsonSchema/JsonSerializerSettings.cs
+++ b/src/NJsonSchema/JsonSerializerSettings.cs
@@ -1,20 +1,32 @@
-﻿namespace NJsonSchema.CodeGeneration.CSharp
+﻿//-----------------------------------------------------------------------
+// <copyright file="JsonSerializerSettings.cs" company="NJsonSchema">
+//     Copyright (c) Rico Suter. All rights reserved.
+// </copyright>
+// <license>https://github.com/rsuter/NJsonSchema/blob/master/LICENSE.md</license>
+// <author>Rico Suter, mail@rsuter.com</author>
+//-----------------------------------------------------------------------
+
+using Newtonsoft.Json;
+
+namespace NJsonSchema.CodeGeneration.CSharp
 {
+    /// <summary>The JSON serializer settings for the generated code.</summary>
     public class JsonSerializerSettings
     {
-        public Newtonsoft.Json.DateParseHandling? DateParseHandling { get; set; }
+        /// <summary>Gets or sets the DateParseHandling settings.</summary>
+        public DateParseHandling? DateParseHandling { get; set; }
 
+        /// <summary>Gets or sets the DateFormatString settings.</summary>
         public string DateFormatString { get; set; }
 
-        override
-        public string ToString() {
-            var settingsString = DateParseHandling == null ? 
-                                    string.Empty : 
-                                    $"DateParseHandling = Newtonsoft.Json.DateParseHandling.{DateParseHandling.ToString()}";
+        /// <summary>Returns a string that represents the current object.</summary>
+        /// <returns>A string that represents the current object.</returns>
+        public override string ToString() {
+            var settingsString = DateParseHandling == null ? string.Empty :
+                $"DateParseHandling = Newtonsoft.Json.DateParseHandling.{DateParseHandling.ToString()}, ";
             
-            settingsString += string.IsNullOrEmpty(DateFormatString) ? 
-                                    string.Empty : 
-                                    $"DateFormatString = \"{DateFormatString}\"";
+            settingsString += string.IsNullOrEmpty(DateFormatString) ? string.Empty : 
+                $"DateFormatString = \"{DateFormatString}\", ";
 
             return settingsString;
         }
--- a/src/NJsonSchema/JsonSerializerSettings.cs
+++ b/src/NJsonSchema/JsonSerializerSettings.cs
@@ -14,7 +14,7 @@
             
             settingsString += string.IsNullOrEmpty(DateFormatString) ? 
                                     string.Empty : 
-                                    $"DateFormatString = {DateFormatString}";
+                                    $"DateFormatString = \"{DateFormatString}\"";
 
             return settingsString;
         }
--- a/src/NJsonSchema/JsonSerializerSettings.cs
+++ b/src/NJsonSchema/JsonSerializerSettings.cs
@@ -10,7 +10,7 @@
         public string ToString() {
             var settingsString = DateParseHandling == null ? 
                                     string.Empty : 
-                                    $"DateParseHandling = {DateParseHandling.ToString()}";
+                                    $"DateParseHandling = Newtonsoft.Json.DateParseHandling.{DateParseHandling.ToString()}";
             
             settingsString += string.IsNullOrEmpty(DateFormatString) ? 
                                     string.Empty : 
--- a/src/NJsonSchema/JsonSerializerSettings.cs
+++ b/src/NJsonSchema/JsonSerializerSettings.cs
@@ -1,0 +1,22 @@
+﻿namespace NJsonSchema.CodeGeneration.CSharp
+{
+    public class JsonSerializerSettings
+    {
+        public Newtonsoft.Json.DateParseHandling? DateParseHandling { get; set; }
+
+        public string DateFormatString { get; set; }
+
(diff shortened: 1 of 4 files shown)


RSuter:
Ref https://github.com/RSuter/NJsonSchema/issues/495

src/NJsonSchema/JsonSerializerSettings.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        override
+        public string ToString() {
+            var settingsString = DateParseHandling == null ? 
+                                    string.Empty : 
+                                    $"DateParseHandling = {DateParseHandling.ToString()}";
+            
+            settingsString += string.IsNullOrEmpty(DateFormatString) ? 
+                                    string.Empty : 
+                                    $"DateFormatString = {DateFormatString}";
```
RSuter: I think we need to escape here with `\"`, eg

      $"DateFormatString = \"{DateFormatString}\"";

src/NJsonSchema/JsonSerializerSettings.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        public Newtonsoft.Json.DateParseHandling? DateParseHandling { get; set; }
+
+        public string DateFormatString { get; set; }
+
+        override
+        public string ToString() {
+            var settingsString = DateParseHandling == null ? 
+                                    string.Empty : 
+                                    $"DateParseHandling = {DateParseHandling.ToString()}";
```
RSuter: I think the namespace and type is missing, eg. 

    $"DateParseHandling = Newtonsoft.Json.DateParseHandling.{DateParseHandling.ToString()}";

lisemon: It is referring to the Property DateParseHandling inside JsonSerializerSettings.

lisemon:
@RSuter Thanks for the quick response 👍  Fixed the stuff.

RSuter:
I have to check the implications of this in the NSwag project before i can merge it... thanks for the pr

RSuter:
I think I'll move the two settings to CSharpGeneratorSettings - to avoid having a collision with the JsonSerializerSettings class of Newtonsoft.Json and because CSharpGeneratorSettings already contains serializer relevant settings... (most of them). Ok for you?

lisemon:
The reason I put the two settings in a separate class is because there are a quite a bunch of serializer settings, that potentially could supported in the long run. But for now I only need the two date specific settings. And as it is not really a generator setting, but more a bundle of serializing setting.
Please have that in mind - but do as you like.

RSuter:
Another completely different option would be to add the setting "JsonSerializerSettingsFactoryMethod" (e.g. "MyNamespace.MyClass.MyStaticMethod" which - when set - gets called in `GenerateJsonSerializerParameterCode()` and you have to implement it somewhere in the project. This way we would be much more flexible and you can configure everything (even converters, switch the contract resolver, etc.). What do you think? This would be a single setting  in CSharpGeneratorSettings...

lisemon:
It is nice to be able to configure a lot 😄  As long as it is well-documented.

RSuter:
Yes, i like the "JsonSerializerSettingsFactoryMethod" setting more, its just one setting which is very flexible, also i only need to add just one new setting to the command line and the UI (see http://nswag.org)

lisemon:
Sounds good 👍

RSuter:
Can you change this? Or should I do it? Id like to release today if possible :-)

lisemon:
Can you do it? I won't be able to look at it until tomorrow (and I'm in a +02:00 timezone)

RSuter:
Ok no problem, i think ill call the setting "JsonSerializerSettingsTransformationMethod" with the signature: 

    JsonSerializerSettings Transform(JsonSerializerSettings settings)

This way we can provide the default settings which can be transformed or completely replaced... sound good?

RSuter:
Released in v9.7, we can still tweak it as long as no one uses it :-)

lisemon:
Awesome with the quick iteration and release!!

I seems to compile fine for serializing, but not for deserializing, as the deserializer does not have the Formatting parameter.

https://www.newtonsoft.com/json/help/html/SerializeObject.htm
https://www.newtonsoft.com/json/help/html/DeserializeObject.htm

E.g.
`csharpGenerator.Settings.JsonSerializerSettingsTransformationMethod = "SerializerSettings.SetupSerializer";`

```
public class SerializerSettings
    {
        public static Formatting SetupSerializer(JsonSerializerSettings jsonSerializerSettings)
        {
            jsonSerializerSettings.DateParseHandling = DateParseHandling.DateTimeOffset;
            return jsonSerializerSettings.Formatting;
        }
    }
}
```

```
public static Visualization FromJson(string data)
        {
            return Newtonsoft.Json.JsonConvert.DeserializeObject<Visualization>(data, SerializerSettings.SetupSerializer(new Newtonsoft.Json.JsonSerializerSettings {  }));
        }

 public string ToJson() 
        {
            return Newtonsoft.Json.JsonConvert.SerializeObject(this, SerializerSettings.SetupSerializer(new Newtonsoft.Json.JsonSerializerSettings {  }));
        }
```

Am I getting something wrong?

lisemon:
Furthermore the setting seems not to be used in the `JsonInheritanceConverter`.

RSuter:
The signatures are: 

    public static T DeserializeObject<T>(string value, JsonSerializerSettings settings)
    public static string SerializeObject(object value, JsonSerializerSettings settings)

**=> Ah sorry, just saw - `SetupSerializer` must return `JsonSerializerSettings`.**

JsonInheritanceConverter: 

Read or write?
For reading we are using the serializer (with hopefully the correct settings): 
https://github.com/RSuter/NJsonSchema/blob/master/src/NJsonSchema/Converters/JsonInheritanceConverter.cs#L120

RSuter:
If you change the SetupSerializer signature, maybe the JsonInheritanceConverter also works correctly?

RSuter:
I've updated the wiki: https://github.com/RSuter/NJsonSchema/wiki/CSharpGeneratorSettings

lisemon:
Ahh yes! 😄 Sorry about that!

I will have to do some testing in order to confirm that it works in the InheritanceConverter 😄

RSuter:
If possible, create a PR with a unit test :-)

RSuter:
... so that it will also not break in the future

lisemon:
Awesome! Maan, you're fast 😄

lisemon:
So at last, I had time to go through with testing... The InheritanceConverter does not use the setting in neither in `ReadJson` or `WriteJson`.

I am mostly concerned with the ReadJson. It is actually the JsonReader that needs the setting. I will have a look at this for the next hour,  and possibly write a unittest.

RSuter:
This is bad... and strange: The InheritanceConverter uses the serializer in both read and write so it should use the given settings...

You can write tests against

https://github.com/RSuter/NJsonSchema/blob/master/src/NJsonSchema/Converters/JsonInheritanceConverter.cs

If everything works, I'll copy this class into the template which is used to generate code.

RSuter:
And we should open a new issue for that...

lisemon:
👍  Jup!

lisemon:
Oh boy I am not even close to having something to push! :D

lisemon:
So I cannot run the test project due to environent issues; as the project type for NJsonSchema.CodeGeneration.Tests is not supported for Visual Studio Community for Mac. I guess it has most todo with the structure of the .csproj-file.

I however note that the `JsonInheritanceConverterTemplateModel` gets the settings in the constructor, and that `JsonInheritanceConverterTemplate.tt` contains the code that is used to generate the .cs-file. I am pretty lost as to how the generation of the .cs-file works but what I can do, it to try and modify the generated model file to make it do as I want, and then we can maybe use that in the .tt-file?

RSuter:
Yes, the test project are not yet .NET Standard - they are legacy MSBuild project using full .NET 4.6. I have to migrate them to .NET Standard and xUnit...

Yes, please change the generated code unit it works and i'll run the tests with it and updated everything...

lisemon:
Thanks. Sorry for not being of more help.

RSuter:
No problem - you are helping... can you try changing the generated code and report here if you've found a solution?

lisemon:
yes!

PR #500, opened 2017-10-04
[WIP] Working towards typescript interfaces having interfaces as children rather than classes

### LINKED ISSUE #494: Typescript client interfaces have full classes as children
If you generate a typescript client and have it generate interfaces for the models, if the interfaces have models as children, those will have the solid types instead of the interface types.
I think these should use the interfaces as their types instead, to make the interfaces more friendly to use.
Will need to fix up some of the other parts that get generated to make this work.

Sound ok?
I’ll work on it later this week if so. Thanks!

danzel:
Fixes #494 eventually.

Dictionaries don't work.
Arrays are a bit hacky (any[], or the second part of TypeScriptTypeResolver.ResolveArray won't be handled correctly)
I'm using tabs in places there should be spaces.

Current output from ClassGenerationTests.When_generating_TypeScript_classes_then_output_is_correct is at:
https://gist.github.com/danzel/7a3cfeafa266c50794fc4f30924327fe

I need a bit of advice on how to change this so Dictionaries can work.

I've added a new property: `NJsonSchema.CodeGeneration.TypeScript.Models.PropertyModel.TypeForInterface`
Which contains the Type that should be used on the interface.
The logic in there is mostly copied from _resolve.Resolve and the methods it calls.
I can get this to work ok for straight objects and arrays (mostly), but for dictionaries the type ends up in the middle of the string, so altering it is a bit harder.

I think I'd like to be able to pass an extra value to `_resolver.Resolve` to specify we are generating an interface and it should add "I" to the front of any class names, but `Resolve` comes from `TypeResolverBase`, so I can't change the definition.

Please take a look over and give me some advice when you can :) Thanks!

### DIFF
--- a/src/NJsonSchema.CodeGeneration.TypeScript/Models/PropertyModel.cs
+++ b/src/NJsonSchema.CodeGeneration.TypeScript/Models/PropertyModel.cs
@@ -49,8 +49,37 @@ namespace NJsonSchema.CodeGeneration.TypeScript.Models
         /// <summary>Gets the type of the property.</summary>
         public override string Type => _resolver.Resolve(_property.ActualPropertySchema, _property.IsNullable(_settings.SchemaType), GetTypeNameHint());
 
+        public string TypeForInterface
+        {
+            get
+            {
+                var schema = _property.ActualPropertySchema.ActualSchema;
+                if (schema.IsDictionary)
+                {
+                    //TODO
+                    return Type + "/*TODO: Dictionary values are not handled*/";
+                }
+                if (IsArray)
+                {
+                    if (schema.Item.ActualSchema.Type.HasFlag(JsonObjectType.Object))
+                    {
+                        return "I" + Type;
+                    }
+                }
+                if (schema.Type.HasFlag(JsonObjectType.Object))
+                {
+                    return "I" + Type;
+                }
+
+                return Type;
+            }
+        }
+
         /// <summary>Gets a value indicating whether the property type is an array.</summary>
         public bool IsArray => _property.ActualPropertySchema.Type.HasFlag(JsonObjectType.Array);
+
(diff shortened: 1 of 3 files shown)


RSuter:
Ok this is a bad one. I think you dont need most of the code because it already exists. Also we cant enhance the resolver with language specific features (interface). Ill look into this today and give you feedback. Youshould wait until then...

RSuter:
I copied your changes to a new branch. I think it looks good, except that one test is broken: 

https://github.com/RSuter/NJsonSchema/tree/typescript-interface-children
https://github.com/RSuter/NJsonSchema/commit/752dfcc228b6840054f9d7b0d2f4b1a2dbd2d289

I think we will not support this (too complicated and very rare): 

            public List<Car[]> Foo { get; set; }

            public Dictionary<string, Skill[]> Bar { get; set; }

Also we need to add more/improve unit tests: 

https://github.com/RSuter/NJsonSchema/blob/typescript-interface-children/src/NJsonSchema.CodeGeneration.Tests/TypeScript/ConstructorInterfaceTests.cs

RSuter:
Can you do more checks on the changes in the branch? This is a major feature which can break things badly... we have to avoid this from happening.

RSuter:
There are two open questions: 

- Should we use `toJSON` check instead of `instanceof`
    - If the object is coming from another "process" (i.e from iframe or popup) the type may be loaded twice and `instanceof` fails. 
    - Checking for toJSON is less generated code
    - This is why i think we shouldnt use instanceof
- Should we put this whole feature behind a setting (default: true)
    - Maybe we introduced a break or a case where not compilable code is generated and this way you can disable the feature until it is fixed
    - This new feature generates much more code which may not be desired (if the feature is not used)

What do you think?

Issue #520, opened 2017-10-18
Typescript generated code for type: 'date' generates bad values


danzel:
I feel bad reporting this because it's my bug! haha :)
Follows #492 #493

The code I added for this is wrong (when your date isn't in utc?)

I'll revisit this tomorrow.
For moment we can use `.format('YYYY-MM-DD')`, will have to figure out Date object.

```js
moment('2017-10-18').toDate()
  Wed Oct 18 2017 00:00:00 GMT+1300 (New Zealand Summer Time)
moment('2017-10-18').toISOString()
  "2017-10-17T11:00:00.000Z"
moment('2017-10-18').format('YYYY-MM-DD')
  "2017-10-18"

moment('2017-10-18 12:00').toISOString()
  "2017-10-17T23:00:00.000Z"
moment('2017-10-18 12:00').toDate()
  Wed Oct 18 2017 12:00:00 GMT+1300 (New Zealand Summer Time)
moment('2017-10-18 12:00').format('YYYY-MM-DD')
  "2017-10-18"

new Date('2017-10-18')
  Wed Oct 18 2017 13:00:00 GMT+1300 (New Zealand Summer Time)
new Date('2017-10-18').toISOString()
  "2017-10-18T00:00:00.000Z"

new Date('2017-10-18 12:00')
  Wed Oct 18 2017 12:00:00 GMT+1300 (New Zealand Summer Time)
new Date('2017-10-18 12:00').toISOString()
  "2017-10-17T23:00:00.000Z"
```

RSuter:
Can you provide a sample?

danzel:
In #493 I made `date` format data work, we produce code like this:

`data["myDate"] = this.myDate ? this.myDate.toISOString().slice(0, 10)`

However, `toISOString` converts the date to UTC, then gets the date portion. Which means we can shift the date (as in the bits I posted above).
We should instead get the year/month/date in whatever timezone the moment/Date object already is in.

RSuter:
So for Date we need: 

    x.getFullYear() + "-" + x.getDate() + "-" + x.getDay()

and for momentJS we need

   m.format("YYYY-MM-DD")

?

danzel:
I think we might need to pad it for `Date`.
But for moment that is correct :+1:

It is meant to be `full-date` format as in https://xml2rfc.tools.ietf.org/public/rfc/html/rfc3339.html#anchor14

danzel:
They should look like:
`2017-01-01`
[...]

PR #528, opened 2017-10-21
Working towards fixing date formatting #520

danzel:
Current NJsonSchema is making all sorts of weird build errors for me (Can't nuget restore DotLiquid 2.0.84 I think?), so I've blind fixed this maybe.

Trouble with fixing it is that the code for formatting a `Date` object can't go all on the right hand side, so I've had to introduce a method and a string replace to plug it in.
Not sure if there is a nicer way to do this?

I also haven't updated the liquid templates or rebuilt the `cs` files from the `tt`.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.Tests/TypeScript/DateCodeGenerationTests.cs
+++ b/src/NJsonSchema.CodeGeneration.Tests/TypeScript/DateCodeGenerationTests.cs
@@ -53,7 +53,7 @@ namespace NJsonSchema.CodeGeneration.Tests.TypeScript
             //// Assert
             Assert.IsTrue(code.Contains("myDate: moment.Moment"));
             Assert.IsTrue(code.Contains("this.myDate = data[\"myDate\"] ? moment(data[\"myDate\"].toString()) : <any>undefined;"));
-            Assert.IsTrue(code.Contains("data[\"myDate\"] = this.myDate ? this.myDate.toISOString().slice(0, 10) : <any>undefined;"));
+            Assert.IsTrue(code.Contains("data[\"myDate\"] = this.myDate ? this.myDate.format('YYYY-MM-DD') : <any>undefined;"));
         }
 
         [TestMethod]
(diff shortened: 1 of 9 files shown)


RSuter:
T4 templates will be removed soon. We need this as liquid template and code artifact.. will wxplain later.

RSuter:
Hi @danzel can you create the fix for the Liquid templates? This replace logic in the template is not possible with Liquid... The T4 templates will be removed soon... For more information: https://github.com/RSuter/NJsonSchema/wiki/Templates

danzel:
Hey, sorry been away a few days.

How do you think I should implement this in liquid?
I could add `DateToStringCodePre` and `DateToStringCodePost`, and put those in?
(`item.{{ DateToStringCode }}` becomes `{{ DateToStringCodePre }}item{{ DateToStringCodePost }}`)
Or can you think of a nicer way?

Need a bit of advice please :) Thanks!

RSuter:
I think we should just add a new property UseJsDate and inline the code like

    {% if UseJsDate %}formatDate({{ value }}){% else %}{{ value }}.format(...){% endif %}

Three times?

danzel:
Haha yeah good idea :+1: Will do this later today

RSuter:
Thank you. And please rename UsesJsDate to UseJsDate to be consistent with the rest

RSuter:
And for now, please do all changes in T4 and liquid. I will remove T4 soon, but i dont know exactly when...

danzel:
Ok, had a second go at this, tests work locally.
Do we need to have an option to control generating the `formatDate` function? Where would I put that? (is there an example to refer to?)

RSuter:
Thanks for the PR. 
I removed the T4 templates because it is too complicated to maintain both...
The formatDate function is now added here as artifact: 

https://github.com/RSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration.TypeScript/TypeScriptGenerator.cs#L85

This artifact is later inserted into the file...
These artifacts are used to output into multiple files later: https://github.com/RSuter/NJsonSchema/issues/514

RSuter:
Ok, now I've removed all T4 templates in NJS and NSwag... Please test with the latest NSwag CI builds...

danzel:
image
This bit is padded out an extra tab, everything else looks the same or similar enough :+1:

Issue #550, opened 2017-11-02
Description generated with TypeMapper


alb3ric:
Hello :)

I noticed that if i want to force a description in a type mapper, this one is overriden after. Would be great if it was not :)
Maybe we could check if the description is already set and if so to not override it. (Going to do a PR)

PR #551, opened 2017-11-02
Do not overwrite description if previously set

alb3ric:
Related to #550

### DIFF
--- a/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
+++ b/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
@@ -751,8 +751,8 @@ namespace NJsonSchema.Generation
                         dynamic readOnlyAttribute = propertyAttributes.TryGetIfAssignableTo("System.ComponentModel.ReadOnlyAttribute");
                         if (readOnlyAttribute != null)
                             p.IsReadOnly = readOnlyAttribute.IsReadOnly;
-
-                        p.Description = await propertyInfo.GetDescriptionAsync(propertyAttributes).ConfigureAwait(false);
+                        if (p.Description != null)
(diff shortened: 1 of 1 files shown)


RSuter:
Shouldnt it be

    if (p.Description == null) 

and why is it already set? Where? 

I think the problem here is that the description here is the description of the property not the description of the type... The description of the type should be set directly on the type schema itself...

RSuter:
Btw: The tests fail..

alb3ric:
oops sorry my mistake...

You are right it is : if (p.Description == null) 

It is already set in my TypeMapper :

 public async Task GenerateSchemaAsync(JsonSchema4 schema, TypeMapperContext context)
        {
            schema.Type = JsonObjectType.String;
            schema.Description = string.Join(", ", Enum.GetNames(this.MappedType));
        }

RSuter:
Isn't the mapped type referenced here? Is this a primitive type map? Can you provide a little bit more context?

alb3ric:
I'm doing this as a workaround for handling EnumFlags. I'm describing it as a string, because I want it to be display like this in the json-editor I'm using. So I did a TypeMapper like this : 

```
 public class FlaggedEnumTypeMapper : ITypeMapper
    {
        public FlaggedEnumTypeMapper(Type mappedType)
        {
            this.MappedType = mappedType;
        }

        public Type MappedType { get; private set; }

        public bool UseReference => false;

#pragma warning disable 1998
        public async Task GenerateSchemaAsync(JsonSchema4 schema, TypeMapperContext context)
        {
            schema.Type = JsonObjectType.String;
            schema.Description = string.Join(", ", Enum.GetNames(this.MappedType));
        }
    }
```
And I wanted to display in the description the list of possible values of the enum.

RSuter:
9.10.2 soon

alb3ric:
Great, thanks a lot.

PR #572, opened 2017-11-29
Added support for DisplayNameAttribute for schema title

### LINKED ISSUE #571: DisplayNameAttribute Support for Class Titles
First off, awesome library!  Can we add support for `System.ComponentModel.DisplayNameAttribute` that JSON.NET has?  This enables the `title` on the object to be named differently like what `System.ComponentModel.DataAnnotations.DisplayAttribute` does for properties.

kkaung:
This adds a nice support to set schema title by using `System.ComponentModel.DisplayName` which resolves #571.

```cs
[DisplayName("My Cool Component")]
public class MyComponent
{
}
```

### DIFF
--- a/src/NJsonSchema.Tests/Conversion/TypeToSchemaTests.cs
+++ b/src/NJsonSchema.Tests/Conversion/TypeToSchemaTests.cs
@@ -242,7 +242,29 @@ namespace NJsonSchema.Tests.Conversion
             Assert.IsTrue(schema.Definitions.Any(d => d.Key == "MySubtype"));
             Assert.AreEqual(JsonObjectType.String | JsonObjectType.Null, property.ActualSchema.Item.ActualSchema.Properties["Id"].Type);
         }
+        
+        [TestMethod]
+        public async Task When_converting_object_with_display_name_then_schema_title_should_be_set()
+        {
+            var schema = await JsonSchema4.FromTypeAsync<MyType>();
 
+            Assert.AreEqual("My Type", schema.Title);
+        }
+
+        [TestMethod]
+        public async Task When_converting_object_with_empty_display_name_then_fallback_to_reflection()
+        {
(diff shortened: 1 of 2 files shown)


RSuter:
The problem here is that this name is also used as DTO type name and because of this, this PR is a breaking change. This is why I think this attribute should not be considered in DefaultSchemaNameGenerator but we could use it to only generate the title... 

To avoid any problems with existing users, why dont you just implement a custom schema processor and process the attribute?

See https://github.com/RSuter/NJsonSchema/wiki/Schema-Processors

RSuter:
... as you have seen in the current DefaultSchemaNameGenerator implementation, you can also change the name with the JsonSchemaAttribute

kkaung:
In my project, I have 3rd party integrations providing a config object.  I don't know what these configs look like so they will be able to use normal dotnet attributes to add meta so that I could provide a proper configuration page with validation.  The ideal situation is for the integration to use `System.ComponentModel` and `System.ComponentModel.DataAnnotations` attributes to provide the schema I need.  This way they will not need anything else in their lib.

Due to this, `JsonSchemaAttribute` doesn't seem like a viable option.  I can use `ISchemaProcessor` like you suggested but I thought adding the support for this attribute would be a nice feature.  

The one thing I liked about JSON.NET was its usage of `DisplayNameAttribute` for all titles and `DescriptionAttribute` for all descriptions (which this lib has).

```cs
[DisplayName("Credit Cards")]
[Description("Credit card processor configurations")]
public class CreditCardConfig
{
    [DisplayName("Gateway Url")]
    [Description("Url to the credit card gateway")]
    public string Url { get; set; }
}
```

Can you point me to conflicting DTO for my reference?

RSuter:
Just provide an own implementation of DefaultSchemaNameGenerator via the settings. Because this PR would change how the scheme names are generated it will be a huge breaking change for some users...

Issue #575, opened 2017-12-07
Validation fails on enum null values


AlexeiA:
Hello!
I have valid schema which allows nulls as property values:
```javascript
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "prop": {
      "type": ["string", "null"],
      "enum": ["a", "b", null]
    }
  }
}
```
But schema Validate() method agains simple json `{"prop": null}` returns error `{NotInEnumeration: #/prop}`

PR #584, opened 2017-12-13
fix validation of enum null values (#575)

AlexeiA:
Hello! This is the fix for the #575 issue with two new unit tests.
Not sure if the additional return is allowed by code style rules.


RSuter:
Are you sure that the second unit test is correct?

AlexeiA:
Yes, the test is correct, maybe its name is misleading? The test is passed when validation returns 1 error (fails).

RSuter:
I find it a little strange that in the second test null is not allowed... but if this is how json schema validates, it’s fine... :-)

PR #598, opened 2018-01-12
Get underlying type only once outside of a loop

testfirstcoder:


RSuter:
Thanks for the PR (y)

PR #600, opened 2018-01-15
HasStringEnumConverter method refactoring

testfirstcoder:


testfirstcoder:
using assignment is missing

PR #601, opened 2018-01-15
HasStringEnumConverter method refactoring

testfirstcoder:


RSuter:
The usage of this dynamic reflection is to avoid problems with different Newtonsoft.Json versions in the host application... we'll leave this as is to avoid breaking changes for some users.

PR #628, opened 2018-02-22
Use discriminator key from inheritance attribute when serializing

bgorven:
This ensures that the key that is written matches the key that was read.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.Tests/InheritanceSerializationTests.cs
+++ b/src/NJsonSchema.CodeGeneration.Tests/InheritanceSerializationTests.cs
@@ -10,6 +10,10 @@ using NJsonSchema.CodeGeneration.TypeScript;
 using NJsonSchema.Converters;
 using Xunit;
 
+using System.IO;
+using System.Reflection;
+using System.CodeDom.Compiler;
+
 namespace NJsonSchema.CodeGeneration.Tests
 {
     public class Container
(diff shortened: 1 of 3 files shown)


RSuter:
Thanks for the PR!

src/NJsonSchema.CodeGeneration.Tests/InheritanceSerializationTests.cs
```diff
@@ -259,5 +263,76 @@ public void JsonInheritanceConverter_is_thread_safe()
...
+                        .Where(diagnostic => diagnostic.IsWarningAsError || diagnostic.Severity == Microsoft.CodeAnalysis.DiagnosticSeverity.Error)
+                        .Select(d => d.Location.GetLineSpan().StartLinePosition + " - " + d.GetMessage())) + "\n" + code);
+                }
+
+                return Assembly.Load(stream.GetBuffer());
+            }
+#endif
+
+#if NET451
+            var provider = new CSharpCodeProvider();
```
stijnherreman: This needs to be fully qualified as `Microsoft.CSharp.CSharpCodeProvider`, or a `using` needs to be added. The code doesn't build as-is.

RSuter: Thanks for finding this problem

bgorven: I'm sorry I missed that.

Issue #632, opened 2018-02-28
PropertyModel.RangeMaximumValue throw exception when used on an int64 format type


MaxDeg:
The property RangeMaximumValue in the PropertyModel class throw conversion exception if the OpenApi type is int64 and value big enough (9999999999).
CSharpValueGenerator.GetNumericValue try to parse a string in the proper type but the format given to it by RangeMaximumValue and RangeMinimumValue is wrong.

The following code is the one responsible:
https://github.com/RSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration.CSharp/Models/PropertyModel.cs#L139-L142

We need to update it to handle the int64 and float case. If the format is Long we must specify it as "Double" not as "Integer" in order to avoid conversion exception.
```csharp
var format =
                    _property.Format == JsonFormatStrings.Double ||
                    _property.Format == JsonFormatStrings.Float ||
                    _property.Format == JsonFormatStrings.Decimal ||
                    _property.Format == JsonFormatStrings.Long ?
                        JsonFormatStrings.Double : JsonFormatStrings.Integer;
```

I will try to do a pull request today about it.

PR #634, opened 2018-02-28
fix RangeAttribute issue with int64 value

MaxDeg:
bug fix related to #632


RSuter:
I'm not entirely sure if this is correct. But it's fine for me. Thanks for the PR.

PR #645, opened 2018-03-05
implements Record class style

ili:
I'v failed to implement [Record](https://github.com/RSuter/NSwag/issues/1207) class style using only extension templates because setters should be private and this needs to edit `Class.liquid` 

So:
* new Record class style is implemented
* DotLiquid updated (previous version does not support `sort` array filter)
* .editorconfig added for managing indents

### DIFF
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -1,0 +1,10 @@
+; What is EditorConfig? http://editorconfig.org/
+
+root = true
+
+; general settings
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = crlf
+insert_final_newline = false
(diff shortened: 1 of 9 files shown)


RSuter:
Shouldn't we use lower cased (first letter) parameter names?

ili:
In general I would prefer `lowerCamelCase` for parameters. But while looking current implementation I'v stopped on `property.Name` as it is used already for `JsonProperty`.

If you do prefer using `lowerCamelCase` I'll look forward to introduce needed changes.

RSuter:
I think we should add this as built-in filter: 

https://github.com/RSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration/DefaultTemplateFactory.cs#L174

e.g. lowerCamelCase which we can use everywhere and so that we can also in other places (i.e. remove the BlaLower properties in other model classes). This should just call ConversionUtilities.ConvertToLowerCamelCase. What do you think?

ili:
Ok will do it! )

ili:
Done :)

RSuter:
Thank you, i've updated the ctor with the new filter... Why do we render the original constructor?

ili:
By default it is empty, so if user in some case wants to have another constructor, why should we stop him?

RSuter:
image

PR #667, opened 2018-03-26
Added date-time formats for ISO 8601 (may not be complete list)

ellis-knight:
DateTime.TryParse allows for all sorts of date formats to be parsed successfully such as "dd/MM/yyyy" which is contrary to the spec.
See http://json-schema.org/latest/json-schema-validation.html#rfc.section.7.3.1


RSuter:
Thanks for the PR!

PR #683, opened 2018-04-26
Cache the ContractResolver.

### LINKED ISSUE #682: Significant performance regression with JSON serialisation (caching contract resolver)
It appears that this commit de651b24bc7a0d620ecca3dfe5cfe1fa5d1c1afe has introduced a significant performance penalty for JSON serialisation of schemas. Below are the results of a microbenchmark of calling `ToJson()` for a schema for different versions of NJsonSchema:

 Version |     Mean |     Error |    StdDev |
---------|----------|-----------|-----------|
 9.10.45 | 62.74 ms | 1.195 ms  | 1.554 ms  |
 9.10.42 | 62.24 ms | 1.232 ms  | 2.315 ms  |
 9.10.31 | 67.46 ms | 2.606 ms  | 7.685 ms  |
 9.10.30 | 60.64 ms | 1.208 ms  | 1.809 ms  |
 394f3ec | 60.91 ms | 1.203 ms  | 2.836 ms  |
 de651b2 <-- | 67.25 ms | 2.435 ms  | 7.027 ms  | 
 00a48e2 | 906.5 us | 53.07 us  | 154.0 us  |
 2b0f7e6 | 900.9 us | 37.38 us  | 109.6 us  |
 c0699cb | 869.6 us | 38.21 us  | 110.2 us  |
 bda8b16 | 852.6 us | 31.24 us  | 90.14 us  |
 9.10.29 | 838.9 us | 39.87 us  | 115.7 us  |
 9.10.28 | 802.9 us | 20.88 us  | 60.25 us  |
 9.10.27 | 771.0 us | 14.86 us  | 20.84 us  |
 9.10.26 | 796.7 us | 17.24 us  | 26.84 us  |
 9.10.25 | 783.8 us | 15.56 us  | 25.57 us  |
 9.10.20 | 770.6 us | 15.01 us  | 13.30 us  |
 9.10.0  | 775.8 us | 24.16 us  | 70.48 us  |
 9.9.0   | 758.1 us | 21.05 us  | 57.98 us  |
 9.8.0   | 339.5 us | 12.61 us  | 35.77 us  |
 9.7.0   | 1.655 ms | 0.0782 ms | 0.2231 ms |
 9.6.0   | 1.540 ms | 0.0597 ms | 0.1751 ms |
 9.5.0   | 1.533 ms | 0.0526 ms | 0.1518 ms |
 9.4.6   | 1.537 ms | 0.0503 ms | 0.1452 ms |

The time taken jumps by over 60 times!

It appears that the creation of a custom contract resolver has disabled the Newtonsoft.Json contract resolver state caching which improves performance for repeated serialisation and deserialisation runs.

https://github.com/RSuter/NJsonSchema/blob/de651b24bc7a0d620ecca3dfe5cfe1fa5d1c1afe/src/NJsonSchema/JsonSchema4.cs#L195

Ideally this instance should be shared among repeated serialisation and deserialisations for a single program execution.

I'm not sure how best to resolve this, but can think of a couple of potential solutions:

1. Provide an overload of `ToJson()` and `FromJsonAsync()` that accepts a contract resolver such that the client can perform the caching themselves.
2. Cache the contract resolver in a similar fashion to Newtonsoft.Json

After implementing a proof of concept of solution 2., the resulting benchmark was:

 Method |     Mean |    Error |   StdDev |
------- |---------:|---------:|---------:|
 ToJson | 813.4 us | 9.748 us | 8.140 us |

Thus demonstrating that the caching and not the implementation of the new contract resolver is the cause of the performance regression.

Sumo-MBryant:
NjsonSchema.Demo.Performance now runs 5 times faster.

Fixes #682

### DIFF
--- a/src/NJsonSchema/JsonSchema4.cs
+++ b/src/NJsonSchema/JsonSchema4.cs
@@ -24,6 +24,10 @@ namespace NJsonSchema
     /// <summary>A base class for describing a JSON schema. </summary>
     public partial class JsonSchema4 : IDocumentPathProvider
     {
+        private const SchemaType SerializationSchemaType = SchemaType.JsonSchema;
+        private static Lazy<PropertyRenameAndIgnoreSerializerContractResolver> ContractResolver = new Lazy<PropertyRenameAndIgnoreSerializerContractResolver>(
+            () => CreateJsonSerializerContractResolver(SerializationSchemaType));
+
         private IDictionary<string, JsonProperty> _properties;
         private IDictionary<string, JsonSchema4> _patternProperties;
         private IDictionary<string, JsonSchema4> _definitions;
@@ -188,10 +192,7 @@ namespace NJsonSchema
         /// <returns>The JSON Schema.</returns>
         public static async Task<JsonSchema4> FromJsonAsync(string data, string documentPath, Func<JsonSchema4, JsonReferenceResolver> referenceResolverFactory)
         {
-            var schemaType = SchemaType.JsonSchema;
-            var contractResolver = CreateJsonSerializerContractResolver(schemaType);
-
-            return await JsonSchemaSerialization.FromJsonAsync(data, schemaType, documentPath, referenceResolverFactory, contractResolver);
+            return await JsonSchemaSerialization.FromJsonAsync(data, SerializationSchemaType, documentPath, referenceResolverFactory, ContractResolver.Value);
         }
 
         internal static JsonSchema4 FromJsonWithoutReferenceHandling(string data)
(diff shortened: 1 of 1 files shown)


RSuter:
I think we don't need a dictionary as the resolver is always the same..

Sumo-MBryant:
I have now replaced it with a simpler implementation using `Lazy<>`.

RSuter:
Nice, thanks for the PR.

RSuter:
It would be great, if we had some performance tests in the code base so that we can better improve in this regard... Did you already have some of these?

PR #688, opened 2018-04-27
 Resolve another instance of Contract Resolver not being cached. 

Sumo-MBryant:
Also added benchmarking for JSON serialization.

Using `BenchmarkDotNet` the `NJsonSchema.Benchmark` project can be run to find out the performance of serialization and deserialization.

`NBench` tests are also integrated and should be runnable as normal `xUnit` tests.


RSuter:
Thanks for the PR, will look into it soon.

PR #701, opened 2018-05-22
CS records

ili:
Some fixies to records after real life usage:
* inheritance fix
* no more default values for records (bad practice from one side, and see below)
* properties in constructor now arranged in alphabetical order - this makes code more stable after refactoring (cleanup for ex) of original code and adding new properties.

Added:
* CS compilation assert for CS code (using Roslyn) 
* default generated namespace set to `System` (needed to pass compilation assert)


ili:
No problem, that's good! ))

RSuter:
Ok then. Thanks for the pr, will be in annswag release soon.

ili:
Thanks! )

RSuter:
Now live in NSwag, please report any problems... and thanks for the pr! 👍

PR #707, opened 2018-05-29
When a property has a value tag, that tag is used instead of the summary tag

paulomorgado:
According to StyleCop rules [the `summary` of a property must match its accessors](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1623.md "SA1623PropertySummaryDocumentationMustMatchAccessors").

That means that the `summary` for properties must be something like `Gets or sets …`, which is a description of the .NET property and not its value.

On the other hand, [properties must have a `value` tag](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1609.md "SA1609PropertyDocumentationMustHaveValue") [with text](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1610.md "SA1610PropertyDocumentationMustHaveValueText").

Since the `description` in the swagger is the description of the value, it makes more sense to use this tag.

This PR Solves this.

### DIFF
--- a/src/NJsonSchema.Benchmark/SerializationPerformanceTests.cs
+++ b/src/NJsonSchema.Benchmark/SerializationPerformanceTests.cs
@@ -20,9 +20,9 @@ namespace NJsonSchema.Benchmark
         }
 
         [PerfSetup]
-        #pragma warning disable xUnit1013 // Public method should be marked as test
+#pragma warning disable xUnit1013 // Public method should be marked as test
         public void Setup(BenchmarkContext context)
-        #pragma warning restore xUnit1013 // Public method should be marked as test
+#pragma warning restore xUnit1013 // Public method should be marked as test
         {
             _counter = context.GetCounter("Iterations");
         }
@@ -31,6 +31,7 @@ namespace NJsonSchema.Benchmark
         /// Ensure that we can serialise at least 200 times per second (5ms).
         /// </summary>
         [NBenchFact]
+        [Trait("Category", "SkipWhenLiveUnitTesting")]
         [PerfBenchmark(
             Description = "Ensure serialization doesn't take too long",
             NumberOfIterations = 3,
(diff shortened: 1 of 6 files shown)


paulomorgado:
I'm not sure if the reverse is being done, but, if not, point me in the right direction and I'll do it.

RSuter:
See https://github.com/RSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration.CSharp/Templates/Class.liquid#L30

Maybe we should preserve the original one (custom extension) and provide a setting for this (to avoid breaking changes)?

RSuter:
Thanks for the pr! 👍

RSuter:
The reverse?

paulomorgado:
By reverse I was referring to generate `summary` and `value` tags for properties with the correct wording.

paulomorgado:
It probably should be something like:

```
    /// <summary>{{ property.Description.GetPropertySummaryXmlDocumentationFromDescription() | csharpdocs }}</summary>
    /// <value>{{ property.Description.GetPropertyValueXmlDocumentationFromDescription() | csharpdocs }}</value>
```

I'm still not familiar with DotLiquid templating. The text between the double curly braces is C# code? What does `| csharpdocs` mean?

paulomorgado:
Also, now, `GetXmlSummaryAsync` looks a bit strange now for properties, since it can be Value instead of Summary.

paulomorgado:
I tried to generate the correct documentation for properties but I had to add properties to `PropertyModel` for that (https://github.com/RSuter/NJsonSchema/pull/707/files#diff-4383900dc8b7b5c4e2cfe8617cd3ef18R12).

Despite the name, these two tests seem to be testing the same thing:

* `When_enum_has_description_then_csharp_has_xml_comment` - https://github.com/RSuter/NJsonSchema/pull/707/files#diff-c05b8be927a1d3f06dd8526c840a4801R260
* `Task When_property_has_description_then_csharp_has_xml_comment` - https://github.com/RSuter/NJsonSchema/pull/707/files#diff-c05b8be927a1d3f06dd8526c840a4801R295

paulomorgado:
@RSuter, can you take a look at this? What do you think?

PR #713, opened 2018-06-14
Updated to support Forcing ENUM to be bitwise

martinholden-skillsoft:
The OpenApi/Swagger specs don't lend themselves easily to BITWISE ENUMS.

Updated the CSharpGeneratorSettings to add a flag (default:false) to allow creating ENUMs that are BITWISE flags

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp.Tests/CSharpGeneratorTests.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp.Tests/CSharpGeneratorTests.cs
@@ -1596,6 +1596,35 @@ namespace NJsonSchema.CodeGeneration.CSharp.Tests
             AssertCompile(output);
         }
 
+        public class ClassWithExtensionData
+        {
+            public string Foo { get; set; }
+
+            [JsonExtensionData]
+            public IDictionary<string, object> ExtensionData { get; set; }
+        }
+
+        [Fact]
+        public async Task When_class_has_property_with_JsonExtensionDataAttribute_on_property_then_AdditionalProperties_schema_is_set()
+        {
+            //// Arrange
+            var schema = await JsonSchema4.FromTypeAsync<ClassWithExtensionData>(new JsonSchemaGeneratorSettings { SchemaType = SchemaType.OpenApi3 });
+            var json = schema.ToJson();
+
+            var generator = new CSharpGenerator(schema, new CSharpGeneratorSettings
+            {
+                ClassStyle = CSharpClassStyle.Poco
+            });
+
+            //// Act
+            var output = generator.GenerateFile("PersonAddress");
+
+            //// Assert
+            Assert.Equal(1, schema.ActualProperties.Count);
+            Assert.True(schema.AllowAdditionalProperties);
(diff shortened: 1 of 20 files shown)


RSuter:
Thanks for the PR.
 
Having a global option is nice, but I really want to solve this also on a per enum basis somehow... what do you think about that?

Ref: 

- https://github.com/RSuter/NJsonSchema/issues/226
- https://github.com/RSuter/NSwag/issues/657

martinholden-skillsoft:
Agreed a better solution that is driven by the schema would be best but like the expert (you) I couldn't find any agreed upon approach.

RSuter:
What if we rename your setting to EnforceEnumFlags and add a new custom extension x-enumFlags=true so that we can at least support it in the njs toolchain?

RSuter:
- Renamed setting to EnforceFlaggedEnums
- Added IsFlaggedEnumerable to JsonSchema4 => x-enumFlags (consistent with existing x-enumNames property)
- Open question: How to handle this flag in TypeScript?

/cc @martinholden-skillsoft @paulomorgado @simonlaszcz @FranzP  what do you think of these additions/the solution/this PR?

martinholden-skillsoft:
Looks good to me for c#.

I am not great on TypeScript but from what I have read the exact same logic should be fine - the bitwise progression is the important bit (i.e. 1, 2,4,8 etc)

https://basarat.gitbooks.io/typescript/content/docs/enums.html#enums-as-flags

RSuter:
TypeScript: But that's probably only working in the JS "process" - the bigger problem is that the serialization does not work... So it has to be implemented/done manually for flag enums.

How is serialization of flag enums handled in C# with Newtonsoft.Json?

paulomorgado:
How about `EnforceFlagsEnum` instead? They are flags enums, not flagged enums.

RSuter:
Maybe there is a non-C# name for this concept?

paulomorgado:
Bitwise?

martinholden-skillsoft:
BinaryFlags?

RSuter:
BTW: `EnumWithFlags.Bar | EnumWithFlags.Baz` is serialized to `"Bar, Baz"`

RSuter:
Bitwise and Binary sounds a little bit to low-level :-) and EnforceFlagsEnums sounds weird because of the two plurals... would be EnforceFlagEnums ok and correct?

Because of the special and C#/Newtonsoft only serialization of flags enums, I'd strongly recommend against using them anyway, but we will merge this... :-)

martinholden-skillsoft:
+1 from me

RSuter:
@martinholden-skillsoft somehow your merge didnt work very well, we now have too many changes in this PR.

RSuter:
Ok, if we call the custom JSON Schema extension `x-flagEnum` (boolean, default: false)

RSuter:
Updated wiki: https://github.com/RSuter/NJsonSchema/wiki/Enums#flags-enums

martinholden-skillsoft:
@RSuter What do you need me to do?

RSuter:
Ill just merge and see what happens :)

PR #718, opened 2018-06-18
Allow existing string property to be discriminant

### LINKED ISSUE #686: Question: is it possible for existing field to be discriminant?
Hi!

I'm wondering whether it is possible to have existing model field to be a discriminant?
Something like this:

```csharp
[JsonConverter(typeof(JsonInheritanceConverter("type")))]
[KnownType(typeof(ClassA))]
[KnownType(typeof(ClassB))]
public abstract class BaseClass
{
	public string Type {get;set;}
}
```

Currently it fails with `The JSON property 'type' is defined multiple times on type 'BaseClass'.`
Is it possible to avoid that verification?

Thing is that we use this field not only as discriminant, but for other needs as well (with modified discriminant values, though).

Thanks a lot!

Dreamescaper:
Fixes #686

Previously it was not possible to have existing property as discriminant.
With this PR it is allowed if this property is with string type.

### DIFF
--- a/src/NJsonSchema.Tests/Generation/InheritanceTests.cs
+++ b/src/NJsonSchema.Tests/Generation/InheritanceTests.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using System.Runtime.Serialization;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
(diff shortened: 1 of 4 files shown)


RSuter:
Thx for the PR. Doesnt this break serialization with the JsonInheritanceConverter or just ignore the property? Or what exactly is the purpose of this? Reading the discriminator for deserialized objects? Customizing it will probably not work...

Dreamescaper:
>  Reading the discriminator for deserialized objects? 
Exactly.

> Doesnt this break serialization with the JsonInheritanceConverter or just ignore the property? 
Hm. Yeah, I actually missed that, since I mostly cared about deserialization. Will update PR with fix.

RSuter:
BTW: You can just update your branch, no need to create a new PR

Dreamescaper:
sure :)

Updated PR. Along with serialization fix, added several tests for JsonInheritanceConverter.

RSuter:
Updated wiki: https://github.com/RSuter/NJsonSchema/wiki/Inheritance

RSuter:
Thanks for the PR

PR #724, opened 2018-06-22
Add @ symbol to constructor parameters

flipchart:
Fixes issue where parameters are reserved keywords


RSuter:
Is this a braeking change for non reserved params? Or will the @ be removed by the compiler, intellisense etc if its not needed?

flipchart:
Not a breaking change. You can annotate all params with the @ symbol. In IL the symbol without the @ is always used. Tested it with code I have - doesn't even impact named parameters

RSuter:
I think there is a const list of reserved words somewhere in the code based, we could have used this to change only the required params... :-)

flipchart:
Yeah, that would have produced cleaner output. Makes the template more complex though :) Also, the constructor assignment is done with @ symbols regardless of the name, so if you go the route of only prefixing reserved words, it should be done there as well

PR #733, opened 2018-07-04
Use allOf for all schemas in inheritance

### LINKED ISSUE #732: SchemaType.Swagger2 does not generate Swagger2 complaint schema
Specifically, the behavior of [allOf](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#composition-and-inheritance-polymorphism).

When generating a schema for a derived class, NJsonSchema will define both a list of properties AND a list of allOf (reference to base class). However, **Swagger2 will ignore the list of properties!**  When using swagger tooling/codegen, the result is multiple clones of the base class, all properties of the derived classes are lost, only the name of the class is different.

I actually fault swagger for breaking from JSON Schema spec w/r/t polymorphism. But NJsonSchema advertises this *SchemaType.Swagger2* switch, which *should* work.

When I try to flatten the dependency tree, I get multiple definition of property errors [very similar to this open issue](https://github.com/RSuter/NJsonSchema/issues/489).

The code I am trying to spec out is legacy, therefore I cannot [declare inheritance](https://github.com/RSuter/NJsonSchema/wiki/Inheritance#declare-inheritance). Although I'm not sure declaring inheritance is the issue, since the schema generated in the linked example is likewise not swagger2 compliant.

I have a feeling this is somehow related to the [*$ref* issue in the OpenAPI Spec itself](https://github.com/OAI/OpenAPI-Specification/issues/556).

RSuter:
Closes #732 

Tasks: 

- [x] Implement correct allOf behavior
- [x] Check that code generators still work correctly
- [x] Check dictionary inheritance schema generation


RSuter:
**Dictionary inheritance handling**

```
        public class Foo
        {
            public Bar Bar { get; set; }
        }

        public class Bar : Dictionary<string, string>
        {
            public string Baz { get; set; }
        }
```

***Schema (OpenAPI 3)***

```
{
  "title": "Foo",
  "type": "object",
  "properties": {
    "Bar": {
      "x-nullable": true,
      "oneOf": [
        {
          "$ref": "#/definitions/Bar"
        }
      ]
    }
  },
  "definitions": {
    "Bar": {
      "allOf": [
        {
          "type": "object",
          "additionalProperties": {
            "type": "string"
          }
        },
        {
          "type": "object",
          "properties": {
            "Baz": {
              "type": "string"
            }
          }
        }
      ]
    }
  }
}
```

RSuter:
Question: In the sample above, how can i decide which one of the schemes in allOf is the base schema (i.e. inherited) when both are inline (no schema is referenced)? When one schema is referenced and one is inline, then the referenced one is the base schema... 

Also see my dictionary test in the previous comment...

@darrelmiller Is the generated schema above correct for the given C# code? Can you answer the question in this comment (is there even an answer)?

RSuter:
For reference, there is also oneof inheritance: 
https://github.com/RSuter/NJsonSchema/issues/13

RSuter:
Had to revert, because there are still some issues: New PR: https://github.com/RSuter/NJsonSchema/pull/733

PR #737, opened 2018-07-08
Support for YAML references

Em1ss1oN:
Proposed fix for https://github.com/RSuter/NJsonSchema/issues/736

Injects source transformation function into `JsonReferenceResolver` throug overloads of `JsonSchema4.FromJsonAsync()`, `JsonSchema4.FromFileAsync()` and `JsonSchema4.FromUrlAsync()` methods.


RSuter:
I created a "counter-proposal" based on your branch. This introduces less breaking changes and only uses a "custom" reference resolver... What do you think?

Em1ss1oN:
Thank's for your time. Your solution looks far more simple, bad for me not to take into considerations that extensibility point.

PR #755, opened 2018-08-06
Dictionary key as object for TS v2.1 and higher

popovsergii:
Adds x-dictionaryKey to JSON Schema for dictionaries with an enum key, so that the enum is retained when generating TS or C#. 

Closes https://github.com/RSuter/NSwag/issues/1507

### DIFF
--- a/src/NJsonSchema.CodeGeneration.TypeScript/TypeScriptTypeResolver.cs
+++ b/src/NJsonSchema.CodeGeneration.TypeScript/TypeScriptTypeResolver.cs
@@ -96,11 +96,20 @@ namespace NJsonSchema.CodeGeneration.TypeScript
 
             if (schema.IsDictionary)
             {
+                var defaultType = "string";
+
                 var prefix = addInterfacePrefix &&
                     SupportsConstructorConversion(schema.AdditionalPropertiesSchema) &&
                     schema.AdditionalPropertiesSchema?.ActualSchema.Type.HasFlag(JsonObjectType.Object) == true ? "I" : "";
+
                 var valueType = prefix + ResolveDictionaryValueType(schema, "any", Settings.SchemaType);
-                return $"{{ [key: string] : {valueType}; }}";
+                var keyType = this.Settings.TypeScriptVersion >= 2.1m ? prefix + ResolveDictionaryKeyType(schema, defaultType, Settings.SchemaType): defaultType;
+
+                if (keyType == defaultType) {
+                    return $"{{ [key: {keyType}] : {valueType}; }}";
+                }
+
+                return $"{{ [key in keyof typeof {keyType}] : {valueType}; }}";
             }
 
             return (addInterfacePrefix && SupportsConstructorConversion(schema) ? "I" : "") +
--- a/src/NJsonSchema.CodeGeneration.TypeScript/TypeScriptTypeResolver.cs
+++ b/src/NJsonSchema.CodeGeneration.TypeScript/TypeScriptTypeResolver.cs
@@ -96,24 +96,24 @@ namespace NJsonSchema.CodeGeneration.TypeScript
 
             if (schema.IsDictionary)
             {
-                var defaultType = "string";
-
                 var prefix = addInterfacePrefix &&
                     SupportsConstructorConversion(schema.AdditionalPropertiesSchema) &&
                     schema.AdditionalPropertiesSchema?.ActualSchema.Type.HasFlag(JsonObjectType.Object) == true ? "I" : "";
 
-                var valueType = prefix + ResolveDictionaryValueType(schema, "any", Settings.SchemaType);
-                var keyType = this.Settings.TypeScriptVersion >= 2.1m ? prefix + ResolveDictionaryKeyType(schema, defaultType, Settings.SchemaType): defaultType;
+                var valueType = prefix + ResolveDictionaryValueType(schema, "any");
 
-                if (keyType == defaultType) {
-                    return $"{{ [key: {keyType}] : {valueType}; }}";
+                var defaultType = "string";
+                var keyType = Settings.TypeScriptVersion >= 2.1m ? prefix + ResolveDictionaryKeyType(schema, defaultType) : defaultType;
+                if (keyType != defaultType)
+                {
+                    return $"{{ [key in keyof typeof {keyType}] : {valueType}; }}";
                 }
 
-                return $"{{ [key in keyof typeof {keyType}] : {valueType}; }}";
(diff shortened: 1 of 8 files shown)


RSuter:
Thank you for the PR. 

However: 

- Item (array item type)/Items (tuple item types) cannot be used here because we are describing the dictionary key - JSON Schema does not have a way to describe that, what about "x-key"? http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.4.1
- I think we should only support enum or string keys, otherwise we should fall back to string, because an object key does not make sense in the JSON world. How would you serialize/deserialize that?
- We should also support that in the C# generator

What do you think?

popovsergii:
Completely agree with you:
1. - Yea It what I was thinking that I did it wrong, what is "x-key"?
2. - Agree, will change it.
3. - will add supporting to the C# generator

Or how better to do it?

RSuter:
I already have the changes give me 10 minutes...

RSuter:
Ok, I've commited the changes. 

What do you think? 

x-key is a custom property which is only known to NJsonSchema (other implementations will fall back to string) - because JSON Schema cannot describe "dictionary" keys (it doesnt even really know dictionaries).

RSuter:
I'm wondering if we even have to put this special treatment behind a setting?

popovsergii:
Amazing!
Yea I think it's just supported by NJS, and if TS v2.1 or higher should be fine, but maybe makes sense to add some setting that is turned on by default.

Thank you!

RSuter:
I renamed x-key to x-dictionaryKey because x-key may already be used by some people and I don't want to cause name conflicts... x-dictionaryKey should be safe. I think we don't add a setting until someone needs/requests it...

RSuter:
@paulomorgado Do you think we can merge this? Any objections?

RSuter:
@popovsergii Is the PR ok from your point of view?

popovsergii:
Yeah, it completely solves my problem, I would like to have it merged, waiting for neget update!

Thank you a lot again!

paulomorgado:
LGTM

hakankurcenlivivense:
NSwag UI shows such an enum key as *. The model does not include the enum, right?

RicoSuter:
I dont understand your question

hakankurcenlivivense:
I am using an "`XyzModel`" among my C# models that directly inherit from `Dictionary<SomeEnum, string>` and it contains no other members. Here is how it is displayed in the NSwag UI:


> description:	
> Generic XyzModel class
> < * >:	string

And "`Example Value`" field displays it as follows:
```
"xyzInstance": {
        "additionalProp1": "string",
        "additionalProp2": "string",
        "additionalProp3": "string"
}
```

I would have expected `*` to be "`SomeEnum`". And instead of "`additionalProp1`", name of enum values could be used.

And here is the TypeScript output produced by the code generation tool:

```
export class XyzModel {

    [key: string]: string | any; 

}

export enum SomeEnum {
    Enum1 = "Enum1",
    Enum2 = "Enum2",
}
```

It is nice that `SomeEnum` is included in the TypeScript models but shouldn't the key of `XyzModel` also be of type `SomeEnum`?

RicoSuter:
> I would have expected * to be "SomeEnum". And instead of "additionalProp1", name of enum values could be used.

This feature only works in nswag/njs but swagger ui (the web ui) is an external tool and does not support it.


> It is nice that SomeEnum is included in the TypeScript models but shouldn't the key of XyzModel also be of type SomeEnum?

Yes it should, are you using nswag to also generate the spec or just the code?

hakankurcenlivivense:
Yes I am using Nswag to generate both the spec and the code.

RicoSuter:
Looks like a bug if the x-dictionryKey is generated... please open an issue.

src/NJsonSchema.CodeGeneration.TypeScript/TypeScriptTypeResolver.cs
```diff
@@ -99,12 +99,21 @@ private string Resolve(JsonSchema4 schema, string typeNameHint, bool addInterfac
...
-                var valueType = prefix + ResolveDictionaryValueType(schema, "any", Settings.SchemaType);
-                return $"{{ [key: string] : {valueType}; }}";
+
+                var valueType = prefix + ResolveDictionaryValueType(schema, "any");
+
+                var defaultType = "string";
+                var keyType = Settings.TypeScriptVersion >= 2.1m ? prefix + ResolveDictionaryKeyType(schema, defaultType) : defaultType;
+                if (keyType != defaultType)
+                {
+                    return $"{{ [key in keyof typeof {keyType}] : {valueType}; }}";
```
geirsagberg: The indexer should really be optional, i.e. `[key in keyof typeof {keyType}]?`. Otherwise TS will complain if you create an empty object map, which is a valid dictionary.

RicoSuter: I think the ? is added at another place when the property is not required/nullable. I think this is not the place to handle that.

geirsagberg: How would one mark the index property as not required though?

Here is the class and generated code that is causing me issues:

```csharp
public class AutoChecklistStateModel
{
    public Dictionary<AutoChecklistItem, bool> ItemsCompleted { get; set; } = new Dictionary<AutoChecklistItem, bool>();
}
```

```typescript
export interface AutoChecklistStateModel {
    itemsCompleted?: { [key in keyof typeof AutoChecklistItem]: boolean; } | null;
}

export enum AutoChecklistItem {
    Unknown = 0,
    FeedStockInventory = 1,
    MedicationStockInventory = 2,
    LiceCount = 3,
    FeedSystemFiles = 4,
}
```

geirsagberg: Causing the following error:

image

RicoSuter: Ah yes, you are right. And changing this should not be a breaking change, right? Can you create a PR? I hope adding this is correct, I’m not that into the TS type system to know all consequences.

geirsagberg: I will create a PR. Since the new TS type will be a more relaxed superset of the current type, I can't see a way for it to break anything :)

geirsagberg: https://github.com/RicoSuter/NJsonSchema/pull/1120

PR #757, opened 2018-08-08
Resolve CSharp string types with uri format as System.Uri

poloBBQ:
Support JsonFormatStrings.Uri at CSharpTypeResolver


RSuter:
Looks good to me. Is Newtonsoft.Json able to serialize/deserialize that correctly?

poloBBQ:
Thanks for the quick response.

I just tested a C# client generated with this change using Newtonsoft.Json 11.0.2 calling the ToJson and FromJson methods and it serialized and deserialized correctly.

RSuter:
Added tests, everything looks as expected...

PR #762, opened 2018-08-18
Adds ExportTypes support.

pekspro:


src/NJsonSchema.CodeGeneration.TypeScript/Templates/Class.liquid
```diff
@@ -1,7 +1,7 @@
 ﻿{% if HasDescription -%}
 /** {{ Description }} */
 {% endif -%}
-export {% if IsAbstract %}abstract {% endif %}class {{ ClassName }}{{ Inheritance }} {
+{% if ExportTypes -%}export {% endif -%}{% if IsAbstract %}abstract {% endif %}class {{ ClassName }}{{ Inheritance }} {
```
RSuter: -%} is only needed to remove the following line break. Here we only need %} ...

RSuter:
Thanks for the PR! Everything looks very good (there are even unit tests, awesome!). I’ve added a small comment which would be great to “fix”, then we can merge, release and update nswag.

Issue #711, opened 2018-06-12
oneOf/required validation


ikenne:
Given this schema:
```
{
....
"xyz": {
	"type": "object,
	"properties": {
		"A": {  "type": "string" }
		"B": {  "type": "integer" }
		"C": {  "type": "number"  }
	},
	"required": ["A"],
	"additionalProperties": false,
	"oneOf": [
            {
              "required": ["B"]
            },
			{
              "required": ["C"]
            }			
     ]
 }
... 
}
```
And this JSON:
```
 xyz : 
 {
	"A": "abc"
 }
```
I expected the validation to return an error because one of the required fields (B or C) is not present in the JSON. The idea is to have one of B or C.

However, this is not the case. The Validate method validates this JSON with no error.

PR #790, opened 2018-10-04
Fixed "required" properties bug

ShahneRodgers:
Adding tests to try to reproduce issue #771 showed that only properties in the "properties" section are considered when validating the schema; properties listed only in the required section are ignored.


RSuter:
Thanks for the pr. Can you reference the correct issue?

ShahneRodgers:
Oops, sorry about that. It should've been #711.

PR #791, opened 2018-10-04
Incorrect generation for xmlObject when arrays are serialized with XM…

emilw:
…L serializer.

Closes https://github.com/RSuter/NJsonSchema/issues/777

### DIFF
--- a/src/NJsonSchema/Infrastructure/XmlObjectExtension.cs
+++ b/src/NJsonSchema/Infrastructure/XmlObjectExtension.cs
@@ -29,14 +29,35 @@ namespace NJsonSchema.Infrastructure
                     GenerateXmlObject(xmlTypeAttribute.TypeName, xmlTypeAttribute.Namespace, false, false, schema);
             }
         }
+
+        /// <summary>
+        /// Generate XML object for a JSON Schema definition
+        /// </summary>
+        /// <param name="schema">The JSON Schema</param>
+        /// <param name="type">The array type</param>
+        public static void GenerateXmlObjectForArrayType(this JsonSchema4 schema, Type type)
+        {
+            if (schema.IsArray && schema.ParentSchema == null)
+            {
+                GenerateXmlObject($@"ArrayOf{schema.Item.Xml.Name}", null, true, false, schema);
+            }
+        }
+
         /// <summary>Generates XMLObject structure for an array with primitive types</summary>
         /// <param name="schema">The JSON Schema of the item.</param>
         /// <param name="type">The item type.</param>
         public static void GenerateXmlObjectForItemType(this JsonSchema4 schema, Type type)
         {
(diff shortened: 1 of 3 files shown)


emilw:
@Rsuter, do you have any feedback on this? I have been testing it out end to end with NSwag also to see the full result besides the updated Unit tests

src/NJsonSchema/Infrastructure/XmlObjectExtension.cs
```diff
@@ -29,14 +29,35 @@ public static void GenerateXmlObjectForType(this JsonSchema4 schema, Type type)
...
         }
+
+        /// <summary>
+        /// Generate XML object for a JSON Schema definition
+        /// </summary>
+        /// <param name="schema">The JSON Schema</param>
+        /// <param name="type">The array type</param>
+        public static void GenerateXmlObjectForArrayType(this JsonSchema4 schema, Type type)
+        {
+            if (schema.IsArray && schema.ParentSchema == null)
```
RSuter: Why `schema.ParentSchema == null`? I think it is not completely safe to rely on ParentSchema...

RSuter:
Thanks for the PR. Looks good, but to be honest I dont really know the XML feature... did you ensure that you just fixed array handling and didnt break existing stuff?

src/NJsonSchema/Infrastructure/XmlObjectExtension.cs
emilw: What i am after mainly is to detect if it is the first parameter to the controller or not, e.g. not a DTO property of type Array. I it is, the ArrayOf[Name] should be applied. Do you see any other way to detect that?

emilw:
@RSuter , I have been running it end to end tests with the NJsonSchema incorparated with NSwag and executed our full pipleline and compared the full Swagger spec before and after and the only changes i found is that the ArrayOf* and the Int32 => int seems to work and no other changes.
There were some required parameter changes, but i think you have added that since our last NSwag upgrade....so i feel quite good :-).

emilw:
Great, when do you think it's possible to get out a new version of NSwag with this change included?

src/NJsonSchema/Infrastructure/XmlObjectExtension.cs
RSuter: So it's to detect if it's a direct usage in a parameter vs. part of a schema (e.g. property)? I think for this case, checking the ParentSchema is fine... let me know if you see problems in your app with this approach (I personally don't use xml schemas in any project).

RSuter:
Releasing NJS v9.11.1

src/NJsonSchema/Infrastructure/XmlObjectExtension.cs
emilw: Correct, thats the difference. Earlier I assumed that all arrays were properties on a DTO/Schema which is not correct....

PR #794, opened 2018-10-17
Fix for missing type mapping in SchemaResolver when type is an array

daletskyi:
Currently in `GenerateArray` [method](https://github.com/RSuter/NJsonSchema/blob/b85cad2f4b4532ae24fc07427066d90f291cc18d/src/NJsonSchema/Generation/JsonSchemaGenerator.cs#L436) in `JsonSchemaGenerator` there is no call to`schemaResolver.AddSchema` so type mapping is missing (for ex. [here generated json is missing array-like types](https://github.com/RSuter/NSwag/blob/c2c3ad24fd1ef4e434a4a2fdc7c224c0c6f61771/src/NSwag.Commands/Commands/SwaggerGeneration/TypesToSwaggerCommand.cs#L95)). This PR fixes this problem.

### DIFF
--- a/src/NJsonSchema.Tests/Generation/JsonSchemaGeneratorTests.cs
+++ b/src/NJsonSchema.Tests/Generation/JsonSchemaGeneratorTests.cs
@@ -1,0 +1,32 @@
+﻿using NJsonSchema.Annotations;
+using NJsonSchema.Generation;
+using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


RSuter:
Can you explain your actual scenario where this is needed?

RicoSuter:
Very old PR, closing now.

PR #829, opened 2018-12-02
Interface based inheritance - Discussion

falahati:
This is not actually a pull request. Especially with the Git client forcing me to add an empty commit at the beginning. But in any case, I think it is a good piece of code for future consideration.

It does also break some TypeScript tests (which is expected as this is a C# only change for now) and one or two specific general tests.

The reason behind this proposed change is the fact that interfaces are more flexible for the generated code.  And since Swagger generated schema doesn't differentiate between an `abstract class` and an `interface` we should prefer the latter in code generation.

In addition to treating `abstract class`es as `interface`s, this commit also changes the Swagger Schema generator to do the opposite and provide `interface`s with information required for creating an inheritance line, just as it does with `class`es. With little change, this code should also be capable of supporting multiple interfaces as well.

Unfortunately, due to the fact that the default `KnownTypeAttribute` definition doesn't allow usage on `interface` types, this commit includes a copy of this type named `NJsonKnownTypeAttribute` which also allows for dynamic type recognition when used with NetFramework. This may sound bad and seems like a major change, but considering the fact that `KnowTypeAttribute` is `sealed` and even Microsoft has other similar classes like the `ServiceKnownTypeAttribute` it should be alright.

`Generated_Class_and_Interface_json_and_code_are_equal()` test method checks for these changes and can be used to see what are the result of this commit. So please check it out.


falahati:
@ranouf,

This is a proposed feature to increase the flexibility of the library and doesn't directly solves any problem. Especially with TypeScript as modifications are only applied to C# code generator for now. The purpose of this pull request is to have a conversation to see if going with interfaces instead of abstract classes going to help especially in case of multiple interfaces being active or if this approach helps any of `anyOf`, `allOf`, `oneOf` rules.

Unfortunately, it doesn't especially target your issue with the library; sorry about that.

RicoSuter:
Very old PR, closing now.

Issue #832, opened 2018-12-05
Inheritance Wiki


alb3ric:
Hello,

I'm trying to generate a schema with inheritance.
I try the sample code in the wiki (Animal, Dog) but the schema generated is different from the wiki (9.12.7).
I get the result below. Nno oneOf but mappings instead. Why?

{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "title": "Container",
  "type": "object",
  "additionalProperties": false,
  "properties": {
    "Animal": {
      "$ref": "#/definitions/Animal"
    }
  },
  "definitions": {
    "Animal": {
      "type": "object",
      "discriminator": {
        "propertyName": "discriminator",
        "mapping": {
          "Dog": "#/definitions/Dog"
        }
      },
      "additionalProperties": false,
      "required": [
        "discriminator"
      ],
      "properties": {
        "discriminator": {
          "type": "string"
        },
        "Foo": {
          "type": "string"
        }
      }
    },
    "Dog": {
      "allOf": [
        {
          "$ref": "#/definitions/Animal"
        },
        {
          "type": "object",
          "additionalProperties": false,
          "properties": {
            "Bar": {
              "type": "string"
            }
          }
        }
      ]
    }

Thanks in advance for your help.

Regards

alb3ric

thebestnom:
Im pretty sure thats the spec, no? Have you tried codegen it back to see what would it generate to?

RSuter:
This is as expected: Swagger uses all of with the first element the base schema and the second the schema itself

alb3ric:
Hello,

Ok, thanks for your answers.
I had in my sample a class called "Cat" extending Animal. 
What I would have like as a result is this (see below). Is this against the specs?

{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "title": "Container",
  "type": "object",
  "additionalProperties": false,
  "properties": {
    "Animal": {
      "oneOf": [
        {
          "$ref": "#/definitions/Dog"
        },
        {
          "$ref": "#/definitions/Cat"
        }
      ]
    }
  },
  "definitions": {
    "Animal": {
      "type": "object",
      "discriminator": {
        "propertyName": "discriminator",
        "mapping": {
          "Dog": "#/definitions/Dog",
          "Cat": "#/definitions/Cat"
        }
      },
      "x-abstract": true,
      "additionalProperties": false,
      "required": [
        "discriminator"
      ],
      "properties": {
        "Foo": {
          "type": [
[...]

Issue #302, opened 2017-01-20
Add InheritanceType setting


rsuter:
The problem is that for validation we need oneOf inheritance (ie define the child classes): #13

... and for swagger generation we need allOf inheritance (ie define parent classes)

rsuter:
Also support code generators...

Issue #13, opened 2015-12-19
Support oneOf inheritance


rsuter:
https://raw.githubusercontent.com/fge/sample-json-schemas/master/geojson/geometry.json

LeroyK:
A combination of oneOf with allOf almost seems to work, except that the client code generator uses the first reference from the list of oneOf references to generate (C#) client code. It should select the base schema if all oneOf references contain a single reference to that base schema in their allOf list.

Example schema:

```json
{
  "openapi": "3.0.1",
  "info": {
    "title": "Pet API",
    "version": "2019-11-06"
  },
  "paths": {
    "/pet": {
      "get": {
        "tags": [
          "Pet"
        ],
        "operationId": "GetPet",
        "responses": {
          "200": {
            "description": "OK",
            "content": {
              "application/json": {
                "schema": {
                  "oneOf": [
                    {
                      "$ref": "#/components/schemas/Dog"
                    },
                    {
                      "$ref": "#/components/schemas/Bird"
                    }
                  ],
                  "discriminator": {
                    "propertyName": "discriminator"
                  }
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "Bird": {
        "required": [
          "wingspan"
        ],
        "type": "object",
        "allOf": [
          {
            "$ref": "#/components/schemas/Pet"
          }
        ],
        "properties": {
          "wingspan": {
            "type": "integer",
            "format": "int32"
          }
        },
        "additionalProperties": false
      },
      "Dog": {
        "required": [
          "barks"
        ],
        "type": "object",
        "allOf": [
          {
            "$ref": "#/components/schemas/Pet"
          }
        ],
        "properties": {
          "barks": {
            "type": "boolean"
          }
        },
        "additionalProperties": false
      },
      "Pet": {
        "required": [
          "birthDate",
          "discriminator"
        ],
        "type": "object",
        "properties": {
          "discriminator": {
            "type": "string"
          },
          "name": {
            "type": "string",
            "nullable": true
          },
          "birthDate": {
            "type": "string",
            "format": "date-time"
          }
        },
        "additionalProperties": false,
        "discriminator": {
          "propertyName": "discriminator"
        }
      }
    }
 }
}
```

Generates in (``Dog`` instead of ``Pet``).

```c#
System.Threading.Tasks.Task<Dog> GetPetAsync(System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
```

[...]

PR #839, opened 2018-12-07
Support oneOf inheritance for JsonSchema Generation #832

alb3ric:
Hello,

Just made some modifications to generate a jsonschema with oneOf inheritance. I didn't do anything for CodeGeneration as don't need it. 
For schema generation do you think it is the way to go? And what did you think about when you mention inheritanceMode ? (which kind of mode would you propose?)

### DIFF
--- a/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
+++ b/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
@@ -274,23 +274,37 @@ namespace NJsonSchema.Generation
             var useDirectReference = Settings.AllowReferencesWithProperties ||
                 !JsonConvert.DeserializeObject<JObject>(JsonConvert.SerializeObject(referencingSchema)).Properties().Any(); // TODO: Improve performance
 
-            if (useDirectReference && referencingSchema.OneOf.Count == 0)
+            //TODO : use InheritanceMode ?
+            if (Settings.SchemaType == SchemaType.JsonSchema && referencedSchema.KnownTypesSchemas.Any())
             {
-                referencingSchema.Reference = referencedSchema.ActualSchema;
-            }
-            else if (Settings.SchemaType != SchemaType.Swagger2)
-            {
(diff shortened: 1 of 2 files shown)


RSuter:
Refs: 

- #832
- #13
- #302

paul-michalik:
Any progress on this?

qub1n:
> Any progress on this?

This one is about generating C# OneOf from JsonSchema OneOf.
Is any progress in another direction? I mean generating JsonSchema OneOf from C#  OneOf?

darkflame0:
> > Any progress on this?
> 
> This one is about generating C# OneOf from JsonSchema OneOf. Is any progress in another direction? I mean generating JsonSchema OneOf from C# OneOf?

@RicoSuter Is any progress for it?
I really need it...

icnocop:
Hi.

As a work-around I was able to use an operation processor.

See https://gist.github.com/icnocop/10de946939e5046190219cc8817356c4

It works for my case, but may not be the best solution, and may not support all scenarios.

For example, it may have been better to use a schema processor instead.

Usage, in Startup.cs, in ConfigureServices:
```
services.AddOpenApiDocument((settings, serviceProvider) =>
{
    settings.FlattenInheritanceHierarchy = true;
    settings.SchemaType = SchemaType.OpenApi3;
    settings.OperationProcessors.Add(new OneOfOperationProcessor());
    settings.AllowReferencesWithProperties = true;
});
```

My types are defined and decorated similar to the following:
```
[KnownType(typeof(ItemA))]
[KnownType(typeof(ItemB))]
[JsonConverter(typeof(JsonInheritanceConverter), "discriminator")]
[JsonInheritance("A", typeof(ItemA))]
[JsonInheritance("B", typeof(ItemB))]
public abstract class Item
{
}

public class ItemA : Item
{
}

public class ItemB : Item
{
}
```

I only tested with types, and generic types with one generic type parameter.

RicoSuter:
Very old PR, closing now. Please open another PR against recent source if this is still an issue.

PR #843, opened 2018-12-13
Fix stackoverflow in GetEnumerableItemType

tlg-bf:
Fixes issue https://github.com/RSuter/NJsonSchema/issues/842

I simply adds a check in `GetEnumerableItemType` when going through the interface of a type to only consider the one that implement `IEnumerable`.


RSuter:
Thanks for the PR. I've added a test. Release (+ NSwag) soon.

RSuter:
v9.13.3

PR #859, opened 2018-12-28
Fix nullability in CS/TS resolvers

RSuter:
Closes https://github.com/RSuter/NSwag/issues/1851


RSuter:
Ref https://github.com/RSuter/NJsonSchema/commit/faf5c944f1925261f2ed37d567079b5fdcf22a45

PR #860, opened 2018-12-28
Deal with backslashes in string enum values

plk:
I assume that this is all that's needed?

Closes https://github.com/RSuter/NSwag/issues/1855

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/Templates/Enum.liquid
+++ b/src/NJsonSchema.CodeGeneration.CSharp/Templates/Enum.liquid
@@ -9,7 +9,7 @@
 {
 {% for enum in Enums -%}
 {%   if IsStringEnum -%}
-    [System.Runtime.Serialization.EnumMember(Value = @"{{ enum.Value }}")]
+    [System.Runtime.Serialization.EnumMember(Value = @"{{ enum.Value | replace: '"', '""' }}")]
 {%   endif -%}
 {% if IsEnumAsBitFlags -%}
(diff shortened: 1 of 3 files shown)


RSuter:
Yes, looks good. What does the @ string verbatim identifier solve?

plk:
Protects the backslashes which end up in there. They need to be preserved correctly as `Value` is used for serialization etc. I will test this in my use case shortly, just to make sure.

RSuter:
Using verbatim string makes sense, but maybe we also need to correctly escape " now?

E.g. 

    [System.Runtime.Serialization.EnumMember(Value = @"{{ enum.Value | "\"", "\"\"" }}")]

(not sure if this works/is allowed)

plk:
Hmm, I suspect we would need to replace:
\" (YAML/JSON double-quote escape) with "" (C# verbatim string double-quote escape).

RSuter:
The Value is already deserialized from a json string to an unescaped c# representation

plk:
Ah, ok, then I'll do a quick test.

plk:
Ok - I have put in some things to deal with double-quotes too. All tests pass and I modified one and added another to check double quote behaviour.

plk:
I could probably test this in my project if I knew which .dll to replace in the NSwag command-line install ...

RSuter:
I don't know if you can just replace the dlls (but probably that works). I think you can just replace the `NJsonSchema.CodeGeneration.CSharp.dll` - the path of the binaries are displayed in the first lines of an CLI execution...

PR #862, opened 2019-01-07
Do not overwrite existing discriminator property

alb3ric:
Hello,

Is it possible to not overwrite the existing discriminator if it exists ? 
I'm suggesting the following modification.

Best regards

### DIFF
--- a/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
+++ b/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
@@ -814,11 +814,14 @@ namespace NJsonSchema.Generation
                     };
 
                     typeSchema.DiscriminatorObject = discriminator;
-                    typeSchema.Properties[discriminatorName] = new JsonProperty
+                    if (existingProperty == null)
                     {
-                        Type = JsonObjectType.String,
-                        IsRequired = true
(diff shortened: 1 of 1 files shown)


RSuter:
Why would you declare the property manually in the DTO class? The problem is that it might not behave as expected...

alb3ric:
Hi,

I'm declaring it in the dto class because i want to specify attributes for the discriminator property.
I'm using the generated schema to display a form with json-editor (https://github.com/json-editor/json-editor). I want the property to have a default value but not being display in the form.

RSuter:
I would prefer to not merge this pr and you use a custom schema processor with the SchemaProcessorAttribute to add your additional stuff, defaults etc to the property

alb3ric:
I'm already using a schema processor to add my additional stuff, but the property is then overwrite by GenerateInheritanceDiscriminator

RSuter:
Aren't schema processors executed **after** the schema is generated?

alb3ric:
Oups....sorry for the mistake...
I have started implementing the properties in my dto, then saw it was overwritten, then i did this code modification and then i used the schema processor....forgetting it was executed after....
Sorry again, I'm closing the PR.

PR #866, opened 2019-01-08
If assembly location is not found, then return current directory path in GetXmlDocumentationPathAsync

ersegun:
In GetXmlDocumentationPathAsync method, assembly.Location or assembly.CodeBase may throw errors or so, in such a case, in the catch block get the current working directory path and return it if the regarding xml file exists.

### DIFF
--- a/src/NJsonSchema/Infrastructure/XmlDocumentationExtensions.cs
+++ b/src/NJsonSchema/Infrastructure/XmlDocumentationExtensions.cs
@@ -516,6 +516,7 @@ namespace NJsonSchema.Infrastructure
 
         private static async Task<string> GetXmlDocumentationPathAsync(dynamic assembly)
         {
+            string path;
             try
             {
                 if (assembly == null)
@@ -528,7 +529,6 @@ namespace NJsonSchema.Infrastructure
                 if (Cache.ContainsKey(assemblyName.FullName))
                     return null;
 
-                string path;
                 if (!string.IsNullOrEmpty(assembly.Location))
                 {
                     var assemblyDirectory = DynamicApis.PathGetDirectoryName((string)assembly.Location);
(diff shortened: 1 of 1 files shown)


src/NJsonSchema/Infrastructure/XmlDocumentationExtensions.cs
```diff
@@ -562,6 +562,11 @@ private static async Task<string> GetXmlDocumentationPathAsync(dynamic assembly)
             }
             catch
             {
+                var currentDirectory = await DynamicApis.DirectoryGetCurrentDirectoryAsync();
+                path = DynamicApis.PathCombine(currentDirectory, "bin\\" + assembly.GetName().Name + ".xml");
```
4ux-nbIx: This won't always work. The application and its assemblies are not always located in the bin folder. It can be anything. The only good fallback for a case when assembly Location and CodeBase are unavailable is current app domain's BaseDirectory. 

I'd suggest suppressing exceptions while trying to access assembly location and code base and always try to obtain app domain base directory, but also in a try/catch block, like this:

``` csharp
try
{
    // try to get assembly location
    // or try to get assembly code base
}
catch
{
    // ignore exceptions, don't return yet
}

try
{
    // try to get app domain base directory
}
catch
{
    // everything failed
    return null;
}
```

src/NJsonSchema/Infrastructure/XmlDocumentationExtensions.cs
```diff
@@ -562,6 +562,11 @@ private static async Task<string> GetXmlDocumentationPathAsync(dynamic assembly)
             }
             catch
             {
+                var currentDirectory = await DynamicApis.DirectoryGetCurrentDirectoryAsync();
```
RSuter: Please move this code into the try{} block. This method must never throw an exception...

ersegun: Added a new try/catch to get current directory path. In case any error happens, it will return null.

ersegun: Added a new try/catch to get current directory path. In case any error happens, it will return null.

ersegun:
By the way, is there a way of getting AppDomain.CurrentDomain.BaseDirectory there ? I couldn't use AppDomain in .Net standard.

Issue #877, opened 2019-01-17
Types included with $ref are duplicated


WilfriedRicken:
When including files with $ref (e.g. "$ref": "./C.json") more than once, types will be duplicated instead of being recognized being the same. When using "definitions" and "$ref": "#/definitions/..." instead, it works.

I included four files A.json...E.json which show this behavior. A.json to D.json include with "$ref": "<path>", E.json is the same with "$ref": "#/definitions"

Additionally, they show a second behavior which is also a problem for us: JSON schemas with no properties are ignored and converted to object. That is not OK in case generated classes have a base class, e.g. when generating with ClassStyle = CSharpClassStyle.Prism. I also included a sample program which shows that behavior.

WilfriedRicken:
A.json:

´´´
{
  "type": "object",
  "required": [
    "signalPaths"
  ],
  "description": "This is the class A",
  "properties": {
    "type": {
      "$ref": "./C.json",
      "description": "This is the type of A"
    },
    "signalPaths": {
      "children": "array",
      "items": {
        "$ref": "./B.json",
        "description": "This is a list of children"
      }
    }
  }
}
´´´

WilfriedRicken:
B.json:

```
{
  "type": "object",
  "description": "This is the class B",
  "properties": {
    "id": {
      "type": "string",
      "description": "This is the ID of B"
    },
    "type": {
      "$ref": "./C.json",
      "description": "This is the type of B"
    },
    "empty": {
      "$ref": "./D.json",
      "description": "This one should be of type D and not object"
    }
  }
}
```

WilfriedRicken:
C.json:

```
{
  "type":"string",
  "description":"Possible types",
  "enum":[
    "First",
    "Second"
  ]
}
```

WilfriedRicken:
D.json:

```
{
  "type": "object",
  "description": "Possible types"
}
```

WilfriedRicken:
E.json:

```
{
  "definitions": {
    "B": {
      "type": "object",
      "description": "This is the class B",
      "properties": {
        "id": {
          "type": "string",
          "description": "This is the ID of B"
        },
        "type": {
          "$ref": "#/definitions/C",
          "description": "This is the type of B"
        },
        "empty": {
          "$ref": "#/definitions/D",
          "description": "This one should be of type D and not object"
        }
      }
    },
    "C": {
      "type": "string",
      "description": "Possible types",
      "enum": [
        "First",
        "Second"
      ]
[...]

PR #878, opened 2019-01-21
Fix for broken file name handling which leads to duplicate $refs

WilfriedRicken:
Fixes issue #877 which appears because file names are not normalized.


RSuter:
Thanks for this well tested PR!

Issue #879, opened 2019-01-22
Classes with no properties are not generated in C# when base class is INotifyPropertyChanged or BindableBase


WilfriedRicken:
Classes with no properties are not generated. This is possibly ok in case the class has no base class, but not if the class has a base class like INotifyPropertyChanged or Prism.MVVM.BindableBase.

RSuter:
This is expected behavior - i.e. if a property is of the "any" (i.e. type object with any properties) it is generated as object (C#) or any (TS)… the simplest solution to fix this is to set additionalProperties = false in the schema...

However we could add an execption to this when the property is defined with a schema reference, i.e. the any type is actually defined as a named type in "definitions"...

RicoSuter:
There are now settings like InlineNamedAny which should be set to false and then this should work, please recheck with v10

PR #880, opened 2019-01-22
Generate empty classes if base class is present. 

WilfriedRicken:
Resolves issue #879. Open point is a failing test for that inheritance / dictionary stuff. Don't know how to handle that really.


RicoSuter:
Very old PR, closing now. Issue closed.

PR #881, opened 2019-01-22
Add a license expression to the nuspec

TravisEz13:
A project I work on uses https://clearlydefined.io to generate a third party notices file and it is having trouble detecting your license type.

They said that this change should help resolve the issue.

Thanks


RSuter:
Thanks for the PR

PR #895, opened 2019-02-06
New setting for Asp.NET Core ActionResult type usage in code generation

optuuttila:
This is related to the NSwag issue 1547 (https://github.com/RSuter/NSwag/issues/1547).
That feature requires a new setting that can be passed in the nswag code generation settings file.


RSuter:
I think this setting should not be in NJsonSchema but in NSwag (in the controller settings), see https://github.com/RSuter/NSwag/pull/1942#discussion_r254588257

optuuttila:
Yes, makes sense. I was trying to search for a correct place where the setting should be in and I picked the wrong one :)

I'll make the change soonish.

PR #904, opened 2019-02-13
Added Replace $ with string.Empty in property name generator (C#)

IvanJosipovic:
Fixes https://github.com/RSuter/NSwag/issues/1950


RSuter:
Thanks for the PR!

PR #910, opened 2019-02-25
Interface inheritance

i-e-b:
Accept 'duplicates' when using both `GenerateAbstractProperties`
and `FlattenInheritanceHierarchy`

In the existing code, setting both these flags causing false duplicate exceptions, meaning chains of interfaces can't be described correctly.

This change accepts the first definition of a field, only when both flags are set.


RicoSuter:
Is this PR still needed?

src/NJsonSchema/Generation/JsonSchemaGenerator.cs
```diff
@@ -915,7 +915,11 @@ private async Task LoadPropertyOrFieldAsync(Newtonsoft.Json.Serialization.JsonPr
 
                 var propertyName = GetPropertyName(property, propertyInfo);
                 if (parentSchema.Properties.ContainsKey(propertyName))
+                {
+                    if (Settings.GenerateAbstractProperties && Settings.FlattenInheritanceHierarchy) return; // accept first instance of the duplicate
```
RicoSuter: I think we should use GetActualFlattenInheritanceHierarchy(parentType) instead of FlattenInheritanceHierarchy

i-e-b: That makes sense. Have updated.

i-e-b:
> Is this PR still needed?

Yes, please. I'm currently using this fork and a related NuGet package to handle a tree of C# interfaces (part of a distributed routing system).

Thanks for this very useful project!

i-e-b:
Closing due to age

RicoSuter:
Sorry for no reaction. Would this still make sense?

i-e-b:
Thanks for the reply. I think it makes sense as a feature, but I've moved on from the project that required it, so I don't need it personally.

PR #916, opened 2019-03-01
Add support for nullable array items (CS, TS)

RSuter:
Closes https://github.com/RSuter/NSwag/issues/1951


RSuter:
Question: How should we name the custom attribute (currently JsonSchemaArrayNullableItemsAttribute)?

/cc @jeremyVignelles

jeremyVignelles:
`ItemsCanBeNullAttribute`? This would at least be semantically close to the existing `CanBeNullAttribute`

RSuter:
Good idea, updated the code.

PR #920, opened 2019-03-05
Update CSharpValueGenerator.cs

WISNI4:
uri and guid default values


RSuter:
Thanks for the PR! Ideally we should also add unit tests for that so that we don't have regressions in the future..

PR #924, opened 2019-03-12
Fixed toJSON() serialization of TypeScript objects with Dictionary property

FineWolf:
In the `ConvertToJavaScript.liquid` template, there was a one character typo that was preventing proper generation of the `toJSON()` method for classes with Dictionary properties.

This corrects this error and allows objects to be properly serialized.


RSuter:
Thank you for fixing this!

PR #925, opened 2019-03-12
Allow up to 7 decimal places for fractional seconds in DateTime values

kkolstad:
`Json.NET` defaults to 7 decimal places when serializing `DateTime` values. It would be nice to accommodate that.

### DIFF
--- a/src/NJsonSchema.Tests/Validation/FormatDateTimeTests.cs
+++ b/src/NJsonSchema.Tests/Validation/FormatDateTimeTests.cs
@@ -66,7 +66,7 @@ namespace NJsonSchema.Tests.Validation
             schema.Type = JsonObjectType.String;
             schema.Format = JsonFormatStrings.DateTime;
 
-            var token = new JValue("2015-01-25T15:43:30Z");
+            var token = new JValue("2015-01-25T15:43:30.1234567Z");
 
             //// Act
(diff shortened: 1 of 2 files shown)


src/NJsonSchema.Tests/Validation/FormatDateTimeTests.cs
```diff
@@ -66,7 +66,7 @@ public void When_format_date_time_with_iso8601_then_validation_succeeds()
             schema.Type = JsonObjectType.String;
             schema.Format = JsonFormatStrings.DateTime;
 
-            var token = new JValue("2015-01-25T15:43:30Z");
+            var token = new JValue("2015-01-25T15:43:30.1234567Z");
```
RSuter: Please add a new test case for the new format so that the current one is still checked and we dont have regressions

kkolstad: Added

RSuter:
Thanks for the PR! So the other format is still supported, correct?

kkolstad:
@RSuter correct, the previous test is still there and passes. Any ISO 8601 formatted datetime will pass with up to 7 decimal places.

RSuter:
Now in latest nuget package version

PR #931, opened 2019-03-19
Fix any inheritance

RicoSuter:
Closes https://github.com/RSuter/NSwag/issues/2028

Just fixing the exception will work but then the class will inherit from "object" and not from the actual base class. 

This PR might fix the referenced issue but break other scenarios...


RicoSuter:
Very old PR, closing now.

PR #934, opened 2019-03-25
Adds support for MinLength and MaxLength attributes

milenapetkova:
* Adds support for MinLength and MaxLength attributes when generating CSharp classes.

The generated class will have the System.ComponentModel.DataAnnotations.MinLength and System.ComponentModel.DataAnnotations.MaxLength added when the minItems and maxItems are specified in the schema.
Example:
```cs
public class Person
{
    [Required]
    public string FirstName { get; set; }

    public string MiddleName { get; set; }

    [Required]
    public string LastName { get; set; }

    public Gender Gender { get; set; }

    [Range(2, 5)]
    public int NumberWithRange { get; set; }

    public DateTime Birthday { get; set; }

    public Company Company { get; set; }

    [MinLength(1)]
    [MaxLength(20)]
    public Collection<Car> Cars { get; set; }
}
```

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp.Tests/GeneralGeneratorTests.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp.Tests/GeneralGeneratorTests.cs
@@ -1161,6 +1161,38 @@ namespace NJsonSchema.CodeGeneration.CSharp.Tests
         }
 
         [Fact]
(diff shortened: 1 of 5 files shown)


RSuter:
Thanks for the PR! Very clean and with unit tests - I like that :-)

milenapetkova:
Thank you, @RSuter! 😃 👍

PR #939, opened 2019-03-30
Convert apostrophe and ampersand to underscore in enum names

DanielSmon:
Also changed colon to convert to an underscore instead of a hyphen, as hyphen is not valid either.

This solves an immediate issue I'm having. However, I think a better long term solution would be to white-list instead of black-list characters.

Fixes https://github.com/RicoSuter/NSwag/issues/1265

### DIFF
--- a/src/NJsonSchema.CodeGeneration/DefaultEnumNameGenerator.cs
+++ b/src/NJsonSchema.CodeGeneration/DefaultEnumNameGenerator.cs
@@ -30,11 +30,13 @@ namespace NJsonSchema.CodeGeneration
             }
 
             return ConversionUtilities.ConvertToUpperCamelCase(name
-                .Replace(":", "-").Replace(@"""", @""), true)
+                .Replace(":", "_").Replace(@"""", @""), true)
                 .Replace(".", "_")
                 .Replace(",", "_")
                 .Replace("#", "_")
+                .Replace("&", "_")
                 .Replace("-", "_")
+                .Replace("'", "_")
                 .Replace("\\", "_");
         }
     }
--- a/src/NJsonSchema.CodeGeneration/DefaultEnumNameGenerator.cs
+++ b/src/NJsonSchema.CodeGeneration/DefaultEnumNameGenerator.cs
@@ -37,6 +37,8 @@ namespace NJsonSchema.CodeGeneration
                 .Replace("&", "_")
                 .Replace("-", "_")
                 .Replace("'", "_")
+                .Replace("(", "_")
+                .Replace(")", "_")
                 .Replace("\\", "_");
         }
     }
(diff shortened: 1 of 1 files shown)


DanielSmon:
Just added brackets as well. Cheers!

src/NJsonSchema.CodeGeneration/DefaultEnumNameGenerator.cs
```diff
@@ -30,11 +30,15 @@ public string Generate(int index, string name, object value, JsonSchema4 schema)
             }
 
             return ConversionUtilities.ConvertToUpperCamelCase(name
-                .Replace(":", "-").Replace(@"""", @""), true)
+                .Replace(":", "_").Replace(@"""", @""), true)
```
RicoSuter: I think its important to keep this "-" (inside ConvertToUpperCamelCase) as for example foo-bar will be transformed to FooBar and foo_bar will probably be transformed to Foo_bar...

RicoSuter: That is why we transform to - and not _ inside of ConvertToUpperCamelCase - to have nicer upper camel case identifiers as output

DanielSmon:
Hey Rico, ah that makes sense. I had a case where the hyphens were coming through to the enum name in C# which is invalid. I’ll take a closer look over the next few days to get to the bottom of it so we can cater for both scenarios. Cheers! Daniel

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

src/NJsonSchema.CodeGeneration/DefaultEnumNameGenerator.cs
```diff
@@ -34,7 +34,11 @@ public string Generate(int index, string name, object value, JsonSchema4 schema)
                 .Replace(".", "_")
                 .Replace(",", "_")
                 .Replace("#", "_")
+                .Replace("&", "_")
                 .Replace("-", "_")
+                .Replace("'", "_")
+                .Replace("(", "_")
+                .Replace(")", "_")
                 .Replace("\\", "_");
         }
```
DanielSmon: Hey @RicoSuter, I have reverted the colon back to a hyphen for now so we can get the other special characters in and close this one off. I'm a bit short on time right now so am not able to investigate or implement a more comprehensive solution.

Do you want to proceed just with these special characters?

RicoSuter: Fine for now. But i agree that a white list would be better. But keeping it as is will yield new issues/PRs here and we can see all strange enum names and might find better solutions to just turn everything into underscores...

PR #940, opened 2019-03-31
[Typescript] Generate inheritance from ICloneable interface

stbychkov:
Hello!

I've removed duplicates of constructor interface implementation and added implementation of `ICloneable` interface, that will be generated in the related [PR](https://github.com/RicoSuter/NSwag/pull/2063)

### DIFF
--- a/src/NJsonSchema.CodeGeneration.TypeScript/Models/ClassTemplateModel.cs
+++ b/src/NJsonSchema.CodeGeneration.TypeScript/Models/ClassTemplateModel.cs
@@ -6,8 +6,9 @@
 // <author>Rico Suter, mail@rsuter.com</author>
 //-----------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


RicoSuter:
Cant you just define this interface in your project and if the object adheres to this interface you can just assign/cast to it (duck typing)? i.e. you dont need to implement it explicitely in the class?

stbychkov:
Yes, you're right and I know about this option. 
I just thought this interface would be useful if you provided the generated `clone` method.

RicoSuter:
Sorry for that, but i really try to avoid adding more complexity (ie settings, more switches, etc) to the templates - and i dont see much added value in this feature... but if you insist? 🙂

stbychkov:
No, I don't insist. It's just a small suggestion, so let's leave templates at that and move on. 
Thank you for your time.

PR #941, opened 2019-04-03
Fixes bad System.XML reference

PJB3005:
This was causing a compile failure on Linux because Linux has a case sensitive file system, and it also made compiler warnings in my project using yours.


RicoSuter:
Thanks for the PR!

PR #958, opened 2019-04-25
Fixed some async methods and added ConfigureAwait(false)

timsol:
Fixed DynamicApis.FileReadAllTextAsync method so it does not swallow exceptions from the inner task also.


RicoSuter:
Thanks for the PR!

PR #961, opened 2019-05-03
Fix inheritance when schema in hierarchy does not have properties

RicoSuter:


StingyJack:
> Closes RicoSuter/NSwag#2151

That one isn't closed

> But one test fails and thus it is a breaking change...

@RicoSuter  A failing test does not always mean a change is breaking. Which test was it? (The appveyor link above doesn't let me see which one)

RicoSuter:
There is also a merge conflict.

RicoSuter:
Updated with master.

john-larson:
@RicoSuter Any idea when this might be merged?

Issue #964, opened 2019-05-10
TypeScript Client: Parse dates for interfaces


jannesiera:
It is a known issue (although not well documented) that parsing dates does not work when generating a typescript client to interfaces.

In large TypeScript interfaces it is often more conventient to work with interfaces, but converting dates manually is a huge downside. So it would be preferable to have both.

Here, I propose a solution to refactor the code that is currently being generated, without the need of any new logic. This way date parsing would be possible for interfaces.

The question then is if this should be a 'new' option. Class/Interface/ParsedInterface, or if this can be retrofitted to the existing options.

Since, in its current iteration, generating interfaces with a date format that is any other than string seems nonsensical I would suggest that parsing of dates could be enabled for settings Interface+Moment or Interface+DateTime. Thus it would not be necessary to create a 'new' output option.

**Proposed changes**

Here I provide some example output as-is from a real-world project, and the proposed _ouput_ as it could be. To me it seems like those changes would be rather trivial to apply in the generating code and no business logic has been changed. In essence the parsing methods on the class were never strictly tied to the class, and could easily be refactored to a static helper. Completely relieving the need for the class instance while keeping the exact same functionality.

Old:

```
export class LogItemDto implements ILogItemDto {
    deviceId?: string | undefined;
    timestampUtc?: moment.Moment | undefined;
    description?: string | undefined;
    eventPayload?: any | undefined;
    id?: string | undefined;

    constructor(data?: ILogItemDto) {
        if (data) {
            for (var property in data) {
                if (data.hasOwnProperty(property))
                    (<any>this)[property] = (<any>data)[property];
            }
        }
    }

    init(data?: any) {
        if (data) {
            this.deviceId = data["deviceId"];
            this.timestampUtc = data["timestampUtc"] ? moment(data["timestampUtc"].toString()) : <any>undefined;
            this.description = data["description"];
            this.eventPayload = data["eventPayload"];
            this.id = data["id"];
        }
    }

    static fromJS(data: any): LogItemDto {
        data = typeof data === 'object' ? data : {};
        let result = new LogItemDto();
        result.init(data);
        return result;
    }

    toJSON(data?: any) {
        data = typeof data === 'object' ? data : {};
        data["deviceId"] = this.deviceId;
        data["timestampUtc"] = this.timestampUtc ? this.timestampUtc.toISOString() : <any>undefined;
        data["description"] = this.description;
        data["eventPayload"] = this.eventPayload;
        data["id"] = this.id;
[...]

PR #966, opened 2019-05-15
Introduce parsing utils for interfaces to support automatic parsing for e.g. dates

jannesiera:
https://github.com/RicoSuter/NJsonSchema/issues/964

### DIFF
--- a/src/NJsonSchema.CodeGeneration.TypeScript/DataConversionParameters.cs
+++ b/src/NJsonSchema.CodeGeneration.TypeScript/DataConversionParameters.cs
@@ -34,5 +34,8 @@ namespace NJsonSchema.CodeGeneration.TypeScript
 
         /// <summary>Gets or sets the settings.</summary>
         public TypeScriptGeneratorSettings Settings { get; set; }
+
+        /// <summary>ParsingUtils used for interfaces.</summary>
+        public bool UseParsingUtils { get; set; }
     }
 }
--- a/src/NJsonSchema.CodeGeneration.TypeScript/Models/PropertyModel.cs
+++ b/src/NJsonSchema.CodeGeneration.TypeScript/Models/PropertyModel.cs
@@ -134,11 +134,26 @@ namespace NJsonSchema.CodeGeneration.TypeScript.Models
                         TypeNameHint = PropertyName,
                         Resolver = _resolver,
                         NullValue = _settings.NullValue,
-                        Settings = _settings
+                        Settings = _settings,
+                        UseParsingUtils = false
                     });
                 }
-
-                return string.Empty;
+                else
+                {
+                    return DataConversionGenerator.RenderConvertToClassCode(new DataConversionParameters
+                    {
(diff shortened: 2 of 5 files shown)


jannesiera:
@RicoSuter what do you think of these code changes? Anything missing?

src/NJsonSchema.CodeGeneration.TypeScript/Templates/Interface.liquid
```diff
@@ -12,4 +12,60 @@
 
     [key: string]: {{ IndexerPropertyValueType }}; 
 {% endif -%}
+}
+
+{% if ExportTypes %}export {% endif %}class {{ ClassName }}ParsingUtils {
```
RicoSuter: Only generate when setting is true

src/NJsonSchema.CodeGeneration.TypeScript/DataConversionParameters.cs
```diff
@@ -34,5 +34,8 @@ public class DataConversionParameters
 
         /// <summary>Gets or sets the settings.</summary>
         public TypeScriptGeneratorSettings Settings { get; set; }
+
+        /// <summary>ParsingUtils used for interfaces.</summary>
+        public bool UseParsingUtils { get; set; }
```
RicoSuter: I’d call this UseInterfaceSerializer

src/NJsonSchema.CodeGeneration.TypeScript/Templates/Interface.liquid
RicoSuter: This is basically a copy of the class logic, right? Cant we move it to a new template and reuse it here and in class?

RicoSuter: Rename to {{ ClassName }}Serializer?

RicoSuter: Rename to {{ ClassName }}Serializer?

src/NJsonSchema.CodeGeneration.TypeScript/DataConversionParameters.cs
RicoSuter: Rename to UseInterfaceSerializer and move into the settings?

src/NJsonSchema.CodeGeneration.TypeScript/Models/PropertyModel.cs
```diff
@@ -134,11 +134,26 @@ public string ConvertToClassCode
                         TypeNameHint = PropertyName,
                         Resolver = _resolver,
                         NullValue = _settings.NullValue,
-                        Settings = _settings
+                        Settings = _settings,
+                        UseParsingUtils = false
+                    });
+                }
+                else
```
RicoSuter: When in settings revert this and use new setting

RicoSuter:
+ add some tests

PR #981, opened 2019-05-27
Always allow AllowAdditionalProperties on object via settings

dzmitry-lahoda:
https://github.com/RicoSuter/NSwag/pull/2183

Changes: 
- Add AlwaysAllowAdditionalObjectProperties setting (default: false)
- CSharp Generator: Generate AdditionalProperties property when AllowAdditionalProperties = true
- Remove discriminator property in JsonInheritanceConverter when an actual property does not exist


RicoSuter:
You know that this will break client/DTO code generation completely? Because this is the only way to differentiate between dictionary and object

src/NJsonSchema.Tests/Generation/AllowAdditionalPropertiesTests.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            public string Name { get; set; }
+        }
+
+        public class Employee : Person
+        {
+            public int Salary { get; set; }
+        }
+
+#if !NET45
```
RicoSuter: Why no net45?

src/NJsonSchema/Generation/JsonSchemaGeneratorSettings.cs
```diff
@@ -80,6 +80,9 @@ public JsonSchemaGeneratorSettings()
         /// <summary>Gets or sets a value indicating whether to generate a description with number to enum name mappings (for integer enums only, default: false).</summary>
         public bool GenerateEnumMappingDescription { get; set; }
 
+        /// <summary>Will set `additionalProperties` on all added <see cref="JsonSchema">schema definitions and references</see>(default: false).</summary>
+        public bool AllowAdditionalProperties { get; set;}
```
RicoSuter: I'd call this AlwaysAllowAdditionalProperties because in the case of a dictionary (see generator) it is always true and cannot be switched off...

RicoSuter: Maybe also add a remark that this will break DTO code generation badly (I think)

src/NJsonSchema.Tests/Generation/AllowAdditionalPropertiesTests.cs
dzmitry-lahoda: i have copied from other tests... will remove if not needed?

src/NJsonSchema/Generation/JsonSchemaGeneratorSettings.cs
dzmitry-lahoda: > 
> 
> You know that this will break client/DTO code generation completely? Because this is the only way to differentiate between dictionary and object

i have run all the tests and non failed. would you be able to provide test?

dzmitry-lahoda: i have tried to make the property nullable, so only of it was defined (set explicitly) than it was used, but that broke some tests (I guess tests which do settings reflection-serialization). 
so nullable could make clear and that option(true or false) is decided by other means until set explicitly. 

I will rename to `AlwaysAllowAdditionalProperties`, will add comment about inability to differentiate generation of C# code from that (not sure - may be will try to write tests for that). 

May be I will think if option at all needed for my case. I.e. prevent this kind of forward compatibility or adding new properties.

src/NJsonSchema.Tests/Generation/AllowAdditionalPropertiesTests.cs
RicoSuter: > will remove if not needed?

If it works then remove it

src/NJsonSchema/Generation/JsonSchemaGeneratorSettings.cs
RicoSuter: If it's called AlwaysAllowAdditionalProperties then it should not be a nullable bool but just a bool.

It's just important to know that when this is enabled then code generation won't work as expected (might make sense if only json schema is required). Maybe it even works with this enabled - i hope we only generate a dictionary when no properties are defined... we should check that and maybe implement it...

Of course the tests do not fail because they test with AlwaysAllowAdditionalProperties = false => this is not a breaking change, it could just yield strange behavior when enabled.

RicoSuter:
I think it should work even if this new setting is enabled:

image

It's only treated as dictionary if there are no properties.

RicoSuter:
Maybe a better name is AllowAdditionalObjectProperties?

dzmitry-lahoda:
So renaming into `AllowAdditionalObjectProperties` and write test about generating types from schema. If test fails - document it on AllowAdditionalObjectProperties proprety. Correct? What else?

RicoSuter:
Should be fine then...

I wonder if it should be AlwaysAllowAdditionalObjectProperties or just AllowAdditionalObjectProperties to make it as clear as possible?

dzmitry-lahoda:
Have use considered for objects which are `"additionalProperties" : true` to genenerate `dynamic` instead of object?

RicoSuter:
> Have use considered for objects which are "additionalProperties" : true to genenerate dynamic instead of object?

Added setting for that.

RicoSuter:
Add

```


        [Argument(Name = "AnyType", IsRequired = false, Description = "The any .NET type (default: 'object').")]
        public string AnyType
        {
            get { return Settings.CSharpGeneratorSettings.AnyType; }
            set { Settings.CSharpGeneratorSettings.AnyType = value; }
        }
```

RicoSuter:
I added some changes, please check and confirm that this PR is ready...

PR #999, opened 2019-06-06
Add capabilites to flatten schema with duplicate properties in hierarchy

### LINKED ISSUE #998: Understanding Flatten Inheritance with duplicate properties
I had a look at the inheritance documentation but I was unable to find a decent example with the following flow of logic. I need to hide the base class property and keep the child property, however the type is different only the name is the same.

*This is a specific issue when using* `FlattenInheritanceHierarchy = true`

For example

```C#
public class Person {
   public List<string> Mapping { get; set; }
}

// This will cause Error: "The JSON property 'Mapping' is defined multiple times on type Teacher"
public class Teacher: Person {
    public List<PersonMapping> Mapping { get; set; }
}

// Will be fine and use List<string> Mapping { get; set; }
public class Officer: Person {}
```

From the examples I've seen [JsonIgnoreAttribute] will be applied to every class that inherits 
from **Person** but I only need it to be applied on 1 or 2 inherited classes to use 

```C#
public List<PersonMapping> Mapping { get; set; }
``` 
and the others should still use 
```C#
public List<string> Mapping { get; set; }
```

Essentially I need a way to have a conditional JsonIgnoreAttirbute or a configuration for *FlattenInheritanceHierarchy* that will not use base properties if names match

lunchboxsushi:
Allow a flattened schema to use parent mapping instead of throwing error if property provided

Closes #998


RicoSuter:
What if we don't introduce this setting and just use this: 

```
var propertyAlreadyExists = parentSchema.Properties.ContainsKey(propertyName);
if (propertyAlreadyExists)
{
    if (Settings.GetActualFlattenInheritanceHierarchy(parentType))
    {
        parentSchema.Properties.Remove(propertyName);
    }
    else if (!schemaResolver.ShouldIgnoreDuplicateSchemaProperties && propertyAlreadyExists)
    {
        throw new InvalidOperationException("The JSON property '" + propertyName + "' is defined multiple times on type '" + parentType.FullName + "'.");
    }
}
```

?

It's not really a breaking change as before it threw an exception and it's quite expected to have duplicates when flatten is enabled...

lunchboxsushi:
So by essentially enhancing the capabilities of the existing setting `FlattenInheritanceHierarchy` to allow duplicates. And if they want the reverse workflow simply adding the `[IgnoreJsonAttribute]` on the parent property will work as intended. Awesome, I'll make the change and add additional test cases to cover the reverse flow as well by using the `[IgnoreJsonAttribute]`. Any other suggestions ?

src/NJsonSchema.Tests/Generation/IgnoreDuplicateInheritanceHierarchyTests.cs
```diff
@@ -0,0 +1,58 @@
...
+                var schema = JsonSchema.FromType(typeof(Teacher), settings);
+                var data = schema.ToJson();
+
+                // if there's a definition it correctly parsed the data
+                Assert.True(schema.Definitions.ContainsKey("Map"));
+
+                // if no errors that's good
+                Assert.True(true);
+            }
+            catch
```
RicoSuter: This try/catch and assert.true(true) is not needed as the test will just fail on exception...

src/NJsonSchema/Generation/JsonSchemaGeneratorSettings.cs
```diff
@@ -66,6 +66,9 @@ public JsonSchemaGeneratorSettings()
         /// <summary>Gets or sets a value indicating whether to flatten the inheritance hierarchy instead of using allOf to describe inheritance (default: false).</summary>
         public bool FlattenInheritanceHierarchy { get; set; }
 
+        /// <summary>Gets or sets a value indicating whether to remove duplicates.</summary>
+        public bool IgnoreInhertianceDuplicates { get; set; }
```
RicoSuter: Not needed with the other chnage

RicoSuter:
Just FYI: GetActualFlattenInheritanceHierarchy is used to check the JsonSchemaFlattenAttribute for the given type because you can also selectively enable flatten for a single type...

RicoSuter:
Maybe you should move the tests to the already existing flatten hierarchy test class + also check whether the property is generated with the correct type

lunchboxsushi:
Just to verify 

> check whether the property is generated with the correct type

you mean the generated schema type after calling `FromType()` in the test ?

src/NJsonSchema.Tests/Generation/IgnoreDuplicateInheritanceHierarchyTests.cs
lunchboxsushi: Ah, of course!

RicoSuter:
Yes, schema.Properties...

src/NJsonSchema/Generation/JsonSchemaGeneratorSettings.cs
lunchboxsushi: Removed property as new implementation works much better

lunchboxsushi:
Let me know if there's any additional changes you would like to see. I really appreciate it!

src/NJsonSchema.Tests/Generation/FlattenInheritanceHierarchyTests.cs
```diff
@@ -155,5 +164,25 @@ public async Task When_JsonSchemaFlattenAttribute_is_used_on_class_then_inherite
             Assert.True(schema.Definitions["B"].Properties.ContainsKey("Bbb"));
             Assert.True(schema.Definitions["B"].Properties.ContainsKey("Ccc"));
         }
+
+
```
RicoSuter: Remove unused line break

src/NJsonSchema/Generation/JsonSchemaGenerator.cs
```diff
@@ -1142,11 +1142,21 @@ private void LoadPropertyOrField(JsonProperty jsonProperty, ContextualMemberInfo
...
+                    {
+                        parentSchema.Properties.Remove(propertyName);
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException("The JSON property '" + propertyName + "' is defined multiple times on type '" + parentType.FullName + "'.");
+                    }
                 }
 
+
```
RicoSuter: Remove unused line break

src/NJsonSchema.Tests/Generation/FlattenInheritanceHierarchyTests.cs
```diff
@@ -155,5 +164,25 @@ public async Task When_JsonSchemaFlattenAttribute_is_used_on_class_then_inherite
...
+            };
+
+            //// Act & Assertion
+            var schema = JsonSchema.FromType(typeof(Teacher), settings);
+            var data = schema.ToJson();
+
+            // if there's a definition it correctly parsed the data
+            Assert.True(schema.Definitions.ContainsKey("Schedule"));
+            Assert.True(schema.Properties["Schedule"].Type.HasFlag(JsonObjectType.Null));
+            Assert.True(schema.Properties["Schedule"].Type.HasFlag(JsonObjectType.Array));
```
RicoSuter: Maybe you should also check whether the Type.Items is of type String or HasReference (class) - just express/test what you expect in the case of Teacher (class) and Person (string), no?

RicoSuter:
Added some minor comments. Otherwise I think it looks fine

src/NJsonSchema.Tests/Generation/FlattenInheritanceHierarchyTests.cs
lunchboxsushi: removed

src/NJsonSchema.Tests/Generation/FlattenInheritanceHierarchyTests.cs
lunchboxsushi: Sorry, but I don't think I follow, when you're referring to `Type.Items` there's no access for `Items` if I try to access it from the statement `schema.Properties["Schedule"].Type`.

The only thing that makes sense to me, is to use `schema.Properties["Schedule"].Item.HasReference` as it will mean this is a class object vs string. I don't know the API well enough to see what the underline schema type is class/string

src/NJsonSchema.Tests/Generation/FlattenInheritanceHierarchyTests.cs
```diff
@@ -11,11 +11,20 @@ public class FlattenInheritanceHierarchyTests
         public class Person
         {
             public string Name { get; set; }
+
```
lunchboxsushi: removing while line

src/NJsonSchema.Tests/Generation/FlattenInheritanceHierarchyTests.cs
RicoSuter: Ah yes it is `schema.Properties["Schedule"].Item.HasReference` vs `Item.Type == string` we should check

lunchboxsushi: Ah I understand now, Would you like me to add 1 more test to specifically test that it's "string" when I do `FromType(typeof(Person))` to verify it works as intended? Or can I extend the test *When_FlattenInheritanceHierarchy_is_enabled_then_all_properties_are_in_one_schema* to include an assertion for `schema.Properties["Schedule"].Item.Type == JsonObjectType.String `

lunchboxsushi: For now extended the current test case to include person as well to test the types. I believe this will give a meaningful test as you described earlier. Let me know if there's any issues with this! I really appreciate it.

RicoSuter:
Lgtm

RicoSuter:
Ready to merge? Do the test succeed?

lunchboxsushi:
Yes test do succeed, no issues from integration. Let me run it for the entire project as I've only scoped it to that class.

lunchboxsushi:
*Update*: Ran all tests in project no issues

RicoSuter:
Ok. There will be a preview package in the preview feed. Do you need a nuget.org release?

lunchboxsushi:
That would be Ideal if it's not too much trouble !

RicoSuter:
v10.0.15 in 20 mins

RicoSuter:
Thanks for the PR

lunchboxsushi:
Thank you for the help, 1st PR I've ever done :D!

RicoSuter:
Cool, keep ‘em coming 😏

PR #1017, opened 2019-06-25
Remove unnecessary empty space from Enum.liquid

kanoshin:
There is a trailing whitespace in enum template. Some code editors are removing trailing spaces on an unintentional Ctrl+Save click causing generated files to change.


src/NJsonSchema.CodeGeneration.Tests/EnumGenerationTests.cs
```diff
@@ -287,23 +287,23 @@ public async Task When_property_is_nullable_and_enum_allows_null_then_no_excepti
...
-���������],
-���������""enum"":[  
-������������""�ndring"",
-������������""Nyoprettelse"",
-������������""Udg�r"",
-������������null
-���������]
-������}
-���}
+   ""type"":""object"",
```
kanoshin: I don't understand the nature of that changes. It seems that something was wrong with encoding.

PR #1035, opened 2019-07-19
Fix formatting of Typescript dictionary types

ljvanschie:
Small change that updates the formatting of dictionary types in Typescript. Removes the space before the colon. So that code generation changes from

`Promise<{ [key: string] : number; }>`

to

`Promise<{ [key: string]: number; }>`

which is more generally accepted and often done by autoformatters.


RicoSuter:
The tests fail, can you fix them?

PR #1036, opened 2019-07-23
Include Id, from the chema to be used in the Property Model and Class model,

eriksimonic:
When we generate the models we need to have the ID of the property and the class.-
I Extended the PropertyModel and ClassTemplateModel to include the Id.
Property respects the inheritance.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/Models/ClassTemplateModel.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp/Models/ClassTemplateModel.cs
@@ -39,6 +39,7 @@ namespace NJsonSchema.CodeGeneration.CSharp.Models
                 .Select(property => new PropertyModel(this, property, _resolver, _settings))
                 .ToArray();
 
+            Id = _schema.Id;
             if (schema.InheritedSchema != null)
             {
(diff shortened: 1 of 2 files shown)


src/NJsonSchema.CodeGeneration.CSharp/Models/PropertyModel.cs
```diff
@@ -33,8 +33,17 @@ public class PropertyModel : PropertyModelBase
             _property = property;
             _settings = settings;
             _resolver = typeResolver;
+            
+            Id = property.Id;
```
RicoSuter: Id = property.Id ?? property.ActualSchema.Id;

Can a property have its own id?
Maybe also use a derived property like the Name?

src/NJsonSchema.CodeGeneration.CSharp/Models/ClassTemplateModel.cs
```diff
@@ -39,6 +39,7 @@ public class ClassTemplateModel : ClassTemplateModelBase
                 .Select(property => new PropertyModel(this, property, _resolver, _settings))
                 .ToArray();
 
+            Id = _schema.Id;
```
RicoSuter: Use a derived/calculated property?

RicoSuter:
Add some tests?

RicoSuter:
What is the schema/property id?

PR #1039, opened 2019-08-09
Inlined primitive types aren't decorate with validation attributes

MaxDeg:
I create this PR as a starting point. It's only a POC.

I detected that primitives types (see simple repro below) are inlined to not create wrapper around them (which is correct behavior). The issue here is that if the open api definition contains validation rules. NSwag do not generate the validation attributes.

The follow OpenApi 3.0 definition show the simplest reproduction of the issue (same issue with OpenApi 2.0).

I started this PR with a fix for this specific case (string maximum length). Is the current behavior the desired one or should we fix it? I would be happy to augment this PR to cover all validations.

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Simple repro of missing validation in case of inlined type

paths:
  /echo:
    post:
      summary: Echo a simple message
      operationId: echo
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/message'
      responses:
        '200':
          description: the message to echo
          content:
            application/json:    
              schema:
                $ref: "#/components/schemas/message"

components:
  schemas:
    message:
      type: object
      required:
        - value
      properties:
        value:
          $ref: "#/components/schemas/string50"
    string50:
      type: string
      maxLength: 50
```

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp.Tests/ValidationAttributesTests.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp.Tests/ValidationAttributesTests.cs
@@ -1,0 +1,45 @@
+﻿using System.Threading.Tasks;
+using Xunit;
+
+namespace NJsonSchema.CodeGeneration.CSharp.Tests
+{
+    public class ValidationAttributesTests
+    {
+        [Fact]
+        public async Task When_string_property_has_maxlength_then_stringlength_attribute_is_rendered_in_Swagger_mode()
+        {
+            //// Arrange
+            const string json = @"{
+                        'type': 'object',
+                        'required': [ 'value' ],
+                        'properties': {
+                            'value': {
+                                '$ref': '#/definitions/string50'
+                            }
+                        },
+                'definitions': {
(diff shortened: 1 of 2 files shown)


MaxDeg:
@RicoSutter did you had the time to have a look? Are my assumptions correct ?

RicoSuter:
Is this PR still WIP or ready?

MaxDeg:
For now I only handle one validation on string. If my corrections seems logic then I will finalize it tomorrow.

________________________________

RicoSuter:
I think it's fine, yes.

Can you use ActualSchema instead of ActualTypeSchema?

RicoSuter:
And I think you dont need to check `_property.MaxLength` and `_property.ActualTypeSchema.MaxLength`, just adding the `ActualSchema` indirection `_property.ActualSchema.MaxLength` should be enough

MaxDeg:
Thanks for the feedback. I've finalized the PR :)

RicoSuter:
Thanks for the PR! v10.0.23

PR #1040, opened 2019-08-09
Added JsonSchemaIgnoreAttribute because Newtonsoft.JsonIgnoreAttribut…

dn32:
Added JsonSchemaIgnoreAttribute because Newtonsoft.JsonIgnoreAttribute impacts API usage and in some cases it is necessary to ignore only the json shema


src/NJsonSchema/NJsonSchema.csproj
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard1.0;netstandard2.0;net40;net45</TargetFrameworks>
     <Description>JSON Schema reader, generator and validator for .NET</Description>
-    <Version>10.0.21</Version>
+    <Version>20.0.21</Version>
```
RicoSuter: Please do not change version

dn32: It was an accident. I've already corrected.

RicoSuter:
I think it would be better to name it JsonSchemaIgnoreAttribute so that it is clear that it only applicable for json schema generation and is aligned with the other attributes

RicoSuter:
Another option would be to use a special contract resolver for schema generation:
https://blog.rsuter.com/advanced-newtonsoft-json-dynamically-rename-or-ignore-properties-without-changing-the-serialized-class/

But this pr makes it much simpler and it’s probably quite common so good to add out of the box

RicoSuter:
Don't we need to check the attribute on a class?
Can you add some unit tests?

tjrobinson:
We are facing a similar problem but have solved it with the following, based on the blog post above. Replace `MyAttribute` with whichever attribute (existing or custom) you want to be ignored in the schema.

```csharp
var jsonSchemaGeneratorSettings = new JsonSchemaGeneratorSettings
{
    SerializerSettings = new JsonSerializerSettings
    {
        ContractResolver = new IgnoreMyAttributeContractResolver()
    }
};
```

```csharp
public class IgnoreMyAttributeContractResolver : DefaultContractResolver
{
    protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
    {
        var property = base.CreateProperty(member, memberSerialization);
        var attributes = property.AttributeProvider.GetAttributes(typeof(MyAttribute), true);
        if (attributes != null && attributes.Any())
        {
            property.ShouldSerialize = i => false;
            property.Ignored = true;
        }
        return property;
    }
}
```

PR #1044, opened 2019-08-20
Add GenerateOptionalPropertiesAsNullable setting

RicoSuter:
We should not fix JsonSchemaProperty class (because the schema is not really nullable) but just enhance the C# generator with an additional setting (to avoid breaking changes).


RicoSuter:
TODO: Add tests

RicoSuter:
Better name for GenerateNullableOptionalProperties?

RicoSuter:
v10.0.23

Issue #1047, opened 2019-08-21
NJsonSchema.Tests project fails to build from command-line


Jehoel:
## Repro steps:

1. Clone this repo at `master` (commit 2231bb83268baf6828bda8dc5f8353ed66187ed0)
2. Open a Visual Studio 2017 Developer Command prompt window
3. `cd $/build`
4. `01_Build.bat` errors (see *`01_Build.bat` build errors* heading below)
5. `02_CreatePackages.bat` errors (see *`02_CreatePackages.bat` build errors* heading below)

## Suggestions

* I suggest suppressing the following compiler warnings in test projects only to reduce irrelevant output messages:
    * `CS1998` ("`async` method lacks `await` operators")
    * `CS1591` ("Missing XML comment for publicly visible type or member")

### `01_Build.bat` build errors:

```
"C:\git\forks\NJsonSchema\src\NJsonSchema.sln" (rebuild target) (1) ->
"C:\git\forks\NJsonSchema\src\NJsonSchema.Tests\NJsonSchema.Tests.csproj" (Rebuild target) (3) ->
"C:\git\forks\NJsonSchema\src\NJsonSchema.Tests\NJsonSchema.Tests.csproj" (Build target) (3:5) ->
(CoreCompile target) ->
  Generation\InheritanceTests.cs(305,20): error CS0121: The call is ambiguous between the following methods or properties: 'Assert.Throws<T>(Action)' and 'Assert.Throws<T>(Func<Task>)' [C:\git\forks\NJsonSchema\src\NJsonSchema.Tests\NJsonSchema.Tests.csproj]


"C:\git\forks\NJsonSchema\src\NJsonSchema.sln" (rebuild target) (1) ->
"C:\git\forks\NJsonSchema\src\NJsonSchema.Tests\NJsonSchema.Tests.csproj" (Rebuild target) (3) ->
"C:\git\forks\NJsonSchema\src\NJsonSchema.Tests\NJsonSchema.Tests.csproj" (Build target) (3:6) ->
  Generation\InheritanceTests.cs(305,20): error CS0121: The call is ambiguous between the following methods or properties: 'Assert.Throws<T>(Action)' and 'Assert.Throws<T>(Func<Task>)' [C:\git\forks\NJsonSchema\src\NJsonSchema.Tests\NJsonSchema.Tests.csproj]


"C:\git\forks\NJsonSchema\src\NJsonSchema.sln" (rebuild target) (1) ->
[...]

PR #1048, opened 2019-08-21
Fix/1047

Jehoel:
Adds mitigations for issue #1047

* Suppresses XML and `async`/`await` compiler warnings in test projects due to the (literally) thousands of spurious warnings.

* Fixes build errors (not warnings) that are raised when I run the command-line `build\01_Build.bat` on my computer.

I'm not able to fix the errors from `02_CreatePackages.bat`, however.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.TypeScript/NJsonSchema.CodeGeneration.TypeScript.csproj
+++ b/src/NJsonSchema.CodeGeneration.TypeScript/NJsonSchema.CodeGeneration.TypeScript.csproj
@@ -5,7 +5,9 @@
     <Version>10.0.22</Version>
     <PackageTags>json schema validation generator .net</PackageTags>
     <Copyright>Copyright © Rico Suter, 2018</Copyright>
+    <!--
     <PackageLicenseUrl>https://github.com/RicoSuter/NJsonSchema/blob/master/LICENSE.md</PackageLicenseUrl>
+    -->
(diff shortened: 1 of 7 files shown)


RicoSuter:
These scripts are not needed anymore, we just use “dotnet build”. See azure-piplines in the repo root. Is this pr needed then?

src/NJsonSchema.CodeGeneration.TypeScript/NJsonSchema.CodeGeneration.TypeScript.csproj
```diff
@@ -5,7 +5,9 @@
     <Version>10.0.22</Version>
     <PackageTags>json schema validation generator .net</PackageTags>
     <Copyright>Copyright © Rico Suter, 2018</Copyright>
+    <!--
```
RicoSuter: Also use license type?

Jehoel: I don't know what the tag is inside `csproj` files, only `nuspec` files.

Jehoel:
> Is this pr needed then?

As it makes the Error List window in VS more readable, and fixes that compile error `src/NJsonSchema.Tests/Generation/InheritanceTests.cs` I like to think so.

RicoSuter:
v10.0.23

PR #1052, opened 2019-08-28
Features/enum name generator special chars

MaxKlotz:
## Related issue
#552 

Implemented new replacement of characters that are not supported as class/enum names.

- removed `string` replacement of special characeters
- added `Regex` repleacement
- [Source ](https://stackoverflow.com/a/950651/8336973) of `Regex` for allowed characters


RicoSuter:
Can you fix the conflict? Maybe we should also add the more dpecific first character check?

MaxKlotz:
Sure! The conflict is fixed and I will add the special case for the first letter.

PR #1057, opened 2019-09-12
default creds are now passed when obtaining OpenAPI specs from urls

adamjones1:
This works for me for my spec protected by WinAuth. (Let me know if I'm merging to the wrong branch here, I wasn't sure which to target sorry!)

This is in reference to this issue in the NSwag project: https://github.com/RicoSuter/NSwag/issues/2352. (I can raise an issue for this here if you like.)


RicoSuter:
Is this supported in all the frameworks (.net 4.0, PCL, .net core, .net 4.7, etc.)?

adamjones1:
From what I can tell `HttpClientHandler` is supported everywhere that `HttpClient` is (which the file already uses) - .Net Framework >=4.5, .Net Standard >= 1.1, .Net Core >= 1.0. I'm not entirely sure how the `net40` build already works but I imagine this would work the same way?

PR #1062, opened 2019-09-24
Sanitize the "+" for enum value generation

kassemsandarusi:
I noticed this when generation an OA3 spec with NSwag. I am generating a client from a spec that has a string value like "application/json+something".


kassemsandarusi:
The failing test doesn't seem relevant, and running locally passes. Is this test flaky?

kassemsandarusi:
I committed to confirm flakiness. Looks like my hunch was right.

RicoSuter:
Thanks for the pr

PR #1071, opened 2019-10-02
Add support for ISO8601 (2019-10-01T14:48:19.000Z) datetime parsing in JsonSchema.FromSampleJson

thrixton:
Given a DateTime formatted as per below, we should be able to infer the `date-time` format correctly if we've overridden `DateParseHandling = DateParseHandling.None` in `JsonConvert.DefaultSettings`.

**Sample DateTime**
`2018-08-03T23:51:56.000Z`
The other option is to explicitly set DateParseHandling when we define our settings.
```cs
// SampleJsonSchemaGenerator.cs
public JsonSchema Generate(string json)
{
    var token = JsonConvert.DeserializeObject<JToken>(json, new JsonSerializerSettings
    {
        DateFormatHandling = DateFormatHandling.IsoDateFormat,
        DateParseHandling = DateParseHandling.DateTime // <<< Option 2 >>>

    });

    var schema = new JsonSchema();
    Generate(token, schema, schema, "Anonymous");
    return schema;
}
```

### DIFF
--- a/src/NJsonSchema.Tests/Generation/SampleJsonSchemaGeneratorTests.cs
+++ b/src/NJsonSchema.Tests/Generation/SampleJsonSchemaGeneratorTests.cs
@@ -1,4 +1,5 @@
-﻿using NJsonSchema.Generation;
+﻿using Newtonsoft.Json;
+using NJsonSchema.Generation;
 using Xunit;
 
 namespace NJsonSchema.Tests.Generation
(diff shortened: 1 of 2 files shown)


RicoSuter:
The tests are failing.

thrixton:
Thanks @RicoSuter, apologies for for causing you work.
When I merged current master ([1a2ce00b3e](https://github.com/RicoSuter/NJsonSchema/commit/1a2ce00b3e1e22e78303d1dfeff84b73a2a25392)), I indeed get a failing stack overflow error test.
It's odd though, it's intermittent (happens most of the time), and never happens when I run that test on its own (so I can't step through it).
Do you have any pointers on what could be causing it?

RicoSuter:
Which test is it?

PR #1081, opened 2019-10-15
remove JsonInheritanceAttribute in abstract class

Goxiaoy:
abstract class can not be instantiate

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/Templates/Class.liquid
+++ b/src/NJsonSchema.CodeGeneration.CSharp/Templates/Class.liquid
@@ -4,7 +4,9 @@
 {% if HasDiscriminator -%}
 [Newtonsoft.Json.JsonConverter(typeof(JsonInheritanceConverter), "{{ Discriminator }}")]
 {% for derivedClass in DerivedClasses -%}
+{% if derivedClass.IsAbstract != true -%}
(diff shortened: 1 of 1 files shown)


RicoSuter:
Are you sure we can just remove it? I think this is needed so that the deserializer knows the possible implementations of the (abstract) base class.

Goxiaoy:
I think so.
Look at the function `ReadJson` in generated class `JsonInheritanceConverter`

```
var discriminator = Newtonsoft.Json.Linq.Extensions.Value<string>(jObject.GetValue(_discriminator));
var subtype = GetObjectSubtype(objectType, discriminator);

var objectContract = serializer.ContractResolver.ResolveContract(subtype) as Newtonsoft.Json.Serialization.JsonObjectContract;
if (objectContract == null || System.Linq.Enumerable.All(objectContract.Properties, p => p.PropertyName != _discriminator))
{
    jObject.Remove(_discriminator);
}

try
{
    _isReading = true;
    return serializer.Deserialize(jObject.CreateReader(), subtype);
}
finally
{
    _isReading = false;
}
```
When subtype is an abstract type, `serializer.Deserialize(jObject.CreateReader(), subtype);` will raise an error because abstract class can not be instantiate.

RicoSuter:
Ok yes, makes sense.

RicoSuter:
Thanks for the PR.

Issue #1082, opened 2019-10-15
Newtonsoft.json package version should be upgraded to larger then 10.0.1


goxiaoy:
in .net core
```
[JsonConverter(typeof(JsonInheritanceConverter), typeof(Animal), "discriminator")]
[KnownType(typeof(Dog))]
public class Animal
{
    public string Foo { get; set; }
}

public class Dog : Animal
{
    public string Bar { get; set; }
}
```

When I try to force deserialize json without discriminator property to Dog class

```
var dogJson = @"{
""Bar"":""Test""
}";
var dog= JsonConvert.DeserializeObject<Dog>(dogJson);
```
Error rasied
```
 Newtonsoft.Json.JsonException : Error creating 'NJsonSchema.Converters.JsonInheritanceConverter'.
    ---- Newtonsoft.Json.JsonException : No matching parameterized constructor found for 'NJsonSchema.Converters.JsonInheritanceConverter'.
```

See 
[https://github.com/JamesNK/Newtonsoft.Json/issues/1013](https://github.com/JamesNK/Newtonsoft.Json/issues/1013)
[ https://github.com/JamesNK/Newtonsoft.Json/pull/1002]( https://github.com/JamesNK/Newtonsoft.Json/pull/1002)

Goxiaoy:
And In C# Generator `JsonInheritanceConverter.liquid ` should be updated. 
In `Class.liquid` line 5.

```
[Newtonsoft.Json.JsonConverter(typeof(JsonInheritanceConverter), "{{ Discriminator }}")]
```
should be
```
[Newtonsoft.Json.JsonConverter(typeof(JsonInheritanceConverter), tyepof({{ ClassName}}),"{{ Discriminator }}")]
```

PR #1083, opened 2019-10-16
Json inheritance converter fix

Goxiaoy:
See #1082

### DIFF
--- a/src/NJsonSchema/Converters/JsonInheritanceConverter.cs
+++ b/src/NJsonSchema/Converters/JsonInheritanceConverter.cs
@@ -163,19 +163,28 @@ namespace NJsonSchema.Converters
                 return null;
             }
 
-            var discriminator = jObject.GetValue(_discriminator).Value<string>();
-            var subtype = GetDiscriminatorType(jObject, objectType, discriminator);
-
-            var objectContract = serializer.ContractResolver.ResolveContract(subtype) as JsonObjectContract;
-            if (objectContract == null || objectContract.Properties.All(p => p.PropertyName != _discriminator))
+            Type deserializeType = null;
+            if (_baseType!=null&&objectType != _baseType)
             {
-                jObject.Remove(_discriminator);
+                deserializeType = objectType;
             }
+            else
+            {
+                var discriminator = jObject.GetValue(_discriminator).Value<string>();
+                var subtype = GetDiscriminatorType(jObject, objectType, discriminator);
 
+                var objectContract = serializer.ContractResolver.ResolveContract(subtype) as JsonObjectContract;
+                if (objectContract == null || objectContract.Properties.All(p => p.PropertyName != _discriminator))
+                {
+                    jObject.Remove(_discriminator);
+                }
+
+                deserializeType = subtype;
+            }
             try
             {
                 _isReading = true;
-                return serializer.Deserialize(jObject.CreateReader(), subtype);
+                return serializer.Deserialize(jObject.CreateReader(), deserializeType);
             }
             finally
             {
--- a/src/NJsonSchema/NJsonSchema.csproj
+++ b/src/NJsonSchema/NJsonSchema.csproj
@@ -29,7 +29,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Namotion.Reflection" Version="1.0.7" />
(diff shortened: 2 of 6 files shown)


src/NJsonSchema/Converters/JsonInheritanceConverter.cs
```diff
@@ -163,19 +163,28 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
                 return null;
             }
 
-            var discriminator = jObject.GetValue(_discriminator).Value<string>();
-            var subtype = GetDiscriminatorType(jObject, objectType, discriminator);
-
-            var objectContract = serializer.ContractResolver.ResolveContract(subtype) as JsonObjectContract;
-            if (objectContract == null || objectContract.Properties.All(p => p.PropertyName != _discriminator))
+            Type deserializeType = null;
+            if (_baseType != null && objectType != _baseType)
```
RicoSuter: This part is quite dangerous I think. What if the objectType is not the baseType but still not the actual desired type (e.g. B in A => B => C)?

src/NJsonSchema/NJsonSchema.csproj
```diff
@@ -29,7 +29,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Namotion.Reflection" Version="1.0.7" />
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
```
RicoSuter: Can't we keep v9 to avoid breaking changes and if someone needs to use a newer version he can add it to its application and "overrule" it?

nitzel: @RicoSuter How does that "overruling" work? Do I need to build NJsonSchema myself with the updated Newtonsoft.Json package?

RicoSuter: You should be able to just install a newer version in your app and it should use this (as you fid). Can you provide the full stack trace.

nitzel: Prerequisites: `OtherProject` requires `Newtonsoft.Json 12.0.2`

#### Given `CurrentProject` requires `Newtonsoft.Json` **9.0.1** and `NJsonSchema` and `OtherProject`
---
Warnings during build:
```
Warning	NU1605	Detected package downgrade: Newtonsoft.Json from 12.0.2 to 9.0.1. Reference the package directly from the project to select a different version. 
 CurrentProject -> OtherProject 1.0.0.0 -> Newtonsoft.Json (>= 12.0.2) 
 CurrentProject -> Newtonsoft.Json (>= 9.0.1)	CurrentProject	C:\CurrentProject\CurrentProject.csproj	1	
```

When running the test which creates an instance of  `OtherProject.Engine`
```
System.TypeInitializationException : The type initializer for 'OtherProject.Engine' threw an exception.
  ----> System.IO.FileLoadException : Could not load file or assembly 'Newtonsoft.Json, Version=12.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
   at OtherProject.Engine..ctor(IWorkspace workspace)
   at CurrentProject.JsonSchemaTests.Test() in C:\CurrentProject\JsonSchemaTests.cs:line 35
--FileLoadException
   at OtherProject.Engine..cctor()
```

#### Given `CurrentProject` requires `Newtonsoft.Json` **12.0.2** and `NJsonSchema` and `OtherProject`

When running the test
```
System.IO.FileLoadException : Could not load file or assembly 'Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
   at CurrentProject.JsonSchemaTests.Test()
```

-------------

### Solution

I figured out I could solve the issue by requiring `Newtonsoft.Json 12.0.2` in `CurrentProject` and adding the following to `CurrentProject..csproj`
```xml
<Project>
  <!-- ... -->
  <PropertyGroup>
    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
    <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
  </PropertyGroup>
</Project
```

PR #1091, opened 2019-11-13
Do not use GetResolvedPropertyName

Nisden:
The default behavior for a JsonProperty in Json.NET is to ignore NameingStrategy unless overrideSpecifiedNames is true.

However due to how NJsonSchema used contractResolver.GetResolvedPropertyName the behavior would not match the expected.

Calling ```contractResolver.GetResolvedPropertyName``` would process ```JsonProperty.Name``` that where already correct making them incorrect if overrideSpecifiedNames was false on the NamingStrategy.

If we re-configure ```DefaultPropertyNameHandling == PropertyNameHandling.SnakeCase``` and ```DefaultPropertyNameHandling == PropertyNameHandling.CamelCase``` to overrideSpecifiedNames they act as they use to.

This allows us to use JsonProperty.Name without breaking compatibility.

### DIFF
--- a/src/NJsonSchema.Tests/Generation/JsonPropertyAttributeTests.cs
+++ b/src/NJsonSchema.Tests/Generation/JsonPropertyAttributeTests.cs
@@ -20,6 +20,28 @@ namespace NJsonSchema.Tests.Generation
         }
 
         [Fact]
+        public async Task When_name_of_JsonPropertyAttribute_is_set_then_it_is_used_as_json_property_name_even_with_contactresolver_that_has_nameing_strategy()
+        {
+            var settings = new NJsonSchema.Generation.JsonSchemaGeneratorSettings();
+            settings.SerializerSettings = new JsonSerializerSettings
+            {
+                ContractResolver = new Newtonsoft.Json.Serialization.DefaultContractResolver
+                {
+                    NamingStrategy = new Newtonsoft.Json.Serialization.CamelCaseNamingStrategy()
+                }
+            };
+
+            //// Arrange
+            var schema = JsonSchema.FromType<JsonPropertyAttributeTests.MyJsonPropertyTestClass>(settings);
+
+            //// Act
+            var property = schema.Properties["NewName"];
+
+            //// Assert
+            Assert.Equal("NewName", property.Name);
+        }
+
+        [Fact]
         public async Task When_required_is_always_in_JsonPropertyAttribute_then_the_property_is_required()
         {
             //// Arrange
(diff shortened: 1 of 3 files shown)


RicoSuter:
Won't this break existing users which manually specify and use the CamelCasePropertyNamesContractResolver?

Nisden:
> 
> 
> Won't this break existing users which manually specify and use the CamelCasePropertyNamesContractResolver?



> 
> 
> Won't this break existing users which manually specify and use the CamelCasePropertyNamesContractResolver?

I don't believe so, as [JsonSchemaGenerater](https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema/Generation/JsonSchemaGenerator.cs#L878) uses ResolveContract from [JsonSchemaGeneratorSettings](https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema/Generation/JsonSchemaGeneratorSettings.cs#L185) to get the JsonProperty in the first place.

Nisden:
By default ```[CamelCasePropertyNamesContractResolver](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/CamelCasePropertyNamesContractResolver.cs#L50)``` has ```OverrideSpecifiedNames``` set for its NamingStrategy. Meaning that it overwrites the casing of JsonProperties.

Therefor this change has no effect on users manually specifying CamelCasePropertyNamesContractResolver.

RicoSuter:
This breaks everything:
https://ci.appveyor.com/project/rsuter/nswag-25x6o/build/tests

Nisden:
Looks like I assumed incorrectly that we would always be getting a JsonProperty.

Will add a fallback to: contextualMember in case JsonProperty is null.

RicoSuter:
already added the fallback (see referenced commit)
is that ok?

Nisden:
Think you need to run the PropertyName thought Settings.ActualContractResolver.GetResolvedPropertyName, otherwise PropertyNameHandling.CamelCase / PropertyNameHandling.SnakeCase wont be applied to PropertyName generated for the WebAPIGenerator.

https://github.com/RicoSuter/NJsonSchema/pull/1091/files#diff-67f7e91275d00612d276ab9f67fabbe1L329-L332

RicoSuter:
Is this ok now?
https://github.com/RicoSuter/NJsonSchema/commit/0225d2a5f8c39dafac372db8a09f3462c7b9cde2

Nisden:
Looks good to me

PR #1096, opened 2019-11-19
CSharp code generator range attribute fixes:

### LINKED ISSUE #1067: Exclusive Maximum and Exclusive Minimum are completely ignored by PropertyModel's RangeMinimumValue and RangeMaximumValue property functionality
As the title suggests, basically, setting exclusive limitations proves fruitless after the code is generated.

### LINKED ISSUE #1068: Number 'format' is ignored when the spec is defined via reference ($ref)
When trying to generate code from a YAML we've received, we were getting Int32 out of range exceptions from the code generator.

It transpires that if the field is defined via reference, the format seems to be ignored. After de-referencing the fields, the issue went away.

However, we need a fix for this as we can't keep editing a contractual YAML to factor out the referencing for the affected fields each time we receive a new version

[NSwag Issue with Number Formats.txt](https://github.com/RicoSuter/NJsonSchema/files/3670945/NSwag.Issue.with.Number.Formats.txt)

corner22:
Fix OverflowExceptions that can occur with minimum and maximum properties. resolves RicoSuter/NSwag#2450

Added support for ExclusiveMinimum and ExclusiveMaximum properties. resolves #1067.  note that support for number types is only partial.

Fixed Format properties being ignored when $ref. resolves #1068

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/Models/PropertyModel.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp/Models/PropertyModel.cs
@@ -33,6 +33,34 @@ namespace NJsonSchema.CodeGeneration.CSharp.Models
             _property = property;
             _settings = settings;
             _resolver = typeResolver;
+
+            if (_property.ActualSchema.Minimum.HasValue || _property.ActualSchema.Maximum.HasValue)
+            {
+                if (string.IsNullOrEmpty(_property.ActualSchema.Format))
+                {
+                    if (_property.ActualSchema.Type == JsonObjectType.Integer)
+                    {
+                        // If an integer with no format is specified and then test Minimum and Maximum values are compatible with int32 and if not then explicitly set format to int64
+                        if (IsOutOfRange<int>(_property.ActualSchema.Minimum.GetValueOrDefault()) || IsOutOfRange<int>(_property.ActualSchema.Maximum.GetValueOrDefault()))
+                        {
+                            _property.ActualSchema.Format = JsonFormatStrings.Long;
+                        }
+                    }
+                }
+            }
+
+            bool IsOutOfRange<T>(decimal num)
+            {
+                try
+                {
+                    System.Convert.ChangeType(num, typeof(T));
+                    return false;
+                }
+                catch (System.OverflowException)
+                {
+                    return true;
+                }
+            }
         }
 
         /// <summary>Gets the name of the property.</summary>
--- a/src/NJsonSchema.CodeGeneration.CSharp/Models/PropertyModel.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp/Models/PropertyModel.cs
@@ -151,24 +151,13 @@ namespace NJsonSchema.CodeGeneration.CSharp.Models
                     }
                 }
 
-                var format =
-                    propertyFormat == JsonFormatStrings.Double ||
-                    propertyFormat == JsonFormatStrings.Float ||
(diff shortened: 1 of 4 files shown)


src/NJsonSchema/JsonSchema.cs
```diff
@@ -890,6 +890,34 @@ private static JsonObjectType ConvertStringToJsonObjectType(string value)
...
+                    Convert.ChangeType(num, typeof(T));
+                    return false;
+                }
+                catch (OverflowException)
+                {
+                    return true;
+                }
+            }
+
+            if (Minimum.HasValue || Maximum.HasValue)
```
corner22: @RicoSuter I'm not sure that this is the right place to put this logic as it is within the JsonSchema realm and is not specific to c# code generation.  My only other idea (but I don't like it too much) is to move this logic into the constructor of the NJsonSchema.CodeGeneration.CSharp.Models.PropertyModel class so that it modifies _property.ActualSchema.Format.  something like this:

```
            bool IsOutOfRange<T>(decimal num)
            {
                try
                {
                    System.Convert.ChangeType(num, typeof(T));
                    return false;
                }
                catch (System.OverflowException)
                {
                    return true;
                }
            }

            if (_property.ActualSchema.Minimum.HasValue || _property.ActualSchema.Maximum.HasValue)
            {
                if (string.IsNullOrEmpty(_property.ActualSchema.Format))
                {
                    if (_property.ActualSchema.Type == JsonObjectType.Integer)
                    {
                        // If an integer with no format is specified and then test Minimum and Maximum values are compatible with int32 and if not then explicitly set format to int64
                        if (IsOutOfRange<int>(_property.ActualSchema.Minimum.GetValueOrDefault()) || IsOutOfRange<int>(_property.ActualSchema.Maximum.GetValueOrDefault()))
                        {
                            _property.ActualSchema.Format = JsonFormatStrings.Long;
                        }
                    }
                }
            }

```

 Your thoughts?

RicoSuter: C# generator specific stuff must be moved into the generator model classes, e.g. in NJsonSchema.Generation or NJsonSchema.Generation.CSharp

corner22: I have moved the code out of JsonSchema. Hopefully you can accept this PR now

src/NJsonSchema.CodeGeneration.CSharp/Models/PropertyModel.cs
```diff
@@ -33,6 +33,34 @@ public class PropertyModel : PropertyModelBase
...
+            if (_property.ActualSchema.Minimum.HasValue || _property.ActualSchema.Maximum.HasValue)
+            {
+                if (string.IsNullOrEmpty(_property.ActualSchema.Format))
+                {
+                    if (_property.ActualSchema.Type == JsonObjectType.Integer)
+                    {
+                        // If an integer with no format is specified and then test Minimum and Maximum values are compatible with int32 and if not then explicitly set format to int64
+                        if (IsOutOfRange<int>(_property.ActualSchema.Minimum.GetValueOrDefault()) || IsOutOfRange<int>(_property.ActualSchema.Maximum.GetValueOrDefault()))
+                        {
+                            _property.ActualSchema.Format = JsonFormatStrings.Long;
```
RicoSuter: Just merged but now i see that you are modifying the schema in the generator! This is a big no-go as a generator should not have side-effects. Now i need to find a way to fix this...

Issue #566, opened 2017-11-15
$ref to external json file


zivkan:
I'd like to load a schema that has $refs to other json files with http urls. Currently NJsonSchema throws exceptions saying it can't resolve those references. Do you have plans to support this scenario?

An example schema is http://datafactories.schema.management.azure.com/schemas/2015-09-01/Microsoft.DataFactory.Pipeline.json. You'll need to download it with a client that doesn't automatically redirect to https or ignore https cert hostname mismatch errors.

RSuter:
HTTP and file $refs should be supported...

RSuter:
Please add a sample how you load the schema.

zivkan:
```
static async Task MainAsync()
{
    var schema = await 
        JsonSchema4.FromUrlAsync(
            "http://datafactories.schema.management.azure.com/schemas/2015-09-01/Microsoft.DataFactory.Pipeline.json");
}
```

This gives me the error:
Unhandled Exception: System.AggregateException: One or more errors occurred. (Could not resolve the JSON path 'http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/UTC' with the full JSON path 'http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/UTC'.) ---> System.InvalidOperationException: Could not resolve the JSON path 'http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/UTC' with the full JSON path 'http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/UTC'. ---> System.InvalidOperationException: Could not resolve the JSON path 'https://schema.management.azure.com/schemas/2015-08-19/Microsoft.Search.json#/resourceDefinitions/searchServices' with the full JSON path 'https://schema.management.azure.com/schemas/2015-08-19/Microsoft.Search.json#/resourceDefinitions/searchServices'. ---> System.InvalidOperationException: Could not resolve the JSON path 'https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression' with the full JSON path 'https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#/definitions/expression'. ---> System.InvalidOperationException: Could not resolve the path '#/definitions/SearchServiceProperties'.
   at NJsonSchema.JsonReferenceResolver.ResolveDocumentReference(Object rootObject, String jsonPath)
[...]

Issue #588, opened 2017-12-20
Can't load Azure schema


rchande:
I'm attempting to generate an object model for the Azure resource manager schema at https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json# and running into an exception with version 9.10.16.

Write the following:
```C#
var schema = await NJsonSchema.JsonSchema4.FromUrlAsync(
                @"https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#");
```

Two InvalidOperationExceptions are thrown:
```C#
System.InvalidOperationException: 'Could not resolve the JSON path 'https://schema.management.azure.com/schemas/2015-08-19/Microsoft.Search.json#/resourceDefinitions/searchServices' with the full JSON path 'https://schema.management.azure.com/schemas/2015-08-19/Microsoft.Search.json#/resourceDefinitions/searchServices'.'
```
```
System.InvalidOperationException: 'Could not resolve the JSON path 'https://schema.management.azure.com/schemas/2015-08-19/Microsoft.Search.json#/resourceDefinitions/searchServices' with the full JSON path 'https://schema.management.azure.com/schemas/2015-08-19/Microsoft.Search.json#/resourceDefinitions/searchServices'.'
```

At 
```
   at NJsonSchema.JsonReferenceResolver.<ResolveUrlReferenceWithAlreadyResolvedCheckAsync>d__11.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NJsonSchema.JsonReferenceResolver.<ResolveReferenceAsync>d__9.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NJsonSchema.JsonReferenceResolver.<ResolveReferenceAsync>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #1098, opened 2019-11-25
Bugfix/external $ref resolving

davidmbillie:
My team has an API conformance testing project that utilizes the NJsonSchema library to read yaml specs and validate response data against expected schemas parsed from the specs. The functionality worked when our API Governance team's specs all had $refs contained in single file; however, they are going in the direction of having $refs to schemas in external files for things such as headers and responses.

We started experiencing the "System.InvalidOperationException: Could not resolve the JSON path..." errors similar to the one mentioned in [issue 566](https://github.com/RicoSuter/NJsonSchema/issues/566) when trying to load an external $ref from a `headers.yaml` file from a base `specification.yaml` file. First, I checked if the `headers.yaml` file was formatted correctly by loading it through the `FromFileAsync` method individually, and it worked. So I debugged through NJsonSchema with the base `specification.yaml`, and found that the external schema was being loaded in as an <string, object> Dictionary instead of a JsonSchema, causing an internal InvalidCastException. Adding a condition to `ResolveDocumentReferenceWithoutDereferencing` in `JsonReferenceResolver.cs` to check if the obj is an IDictionary then converting it to a JsonSchema with `FromJsonWithCurrentSettings` worked to resolve the $refs.

However, after solving that ref issue, another occurred when trying to resolve $refs for schemas in an external `responses.yaml` file. I found a similar problem in [issue 1171](https://github.com/RicoSuter/NSwag/issues/1171), but we received a "Unable to cast object of type 'NJsonSchema.JsonSchema' to type 'NSwag.OpenApiResponse'." The exception was happening in `IJsonReferenceBase.Reference`, where in this case `value` was type JsonSchema and `T`  was type OpenApiResponse. I solved this issue by checking if `value` was type T, and if not, getting the Reference property from `value` and casting that to type T. With these changes, our integration tests pass.


RicoSuter:
Can you add a unit test which reproduces the initial problem?

davidmbillie:
I included a generic version of the problem spec with its external ref files. I also noticed that the response still wasn't loaded correctly in my test, so I edited my condition for setting the reference property in JsonReferenceBase.

davidmbillie:
@RicoSuter Can I request a review for this? I have an internal issue open for it.

RicoSuter:
i’m extremely hesitant to change something on this base functionality (ref handling) - mainly because i’m the guy who has to fix it if it blows up for someone (which happened often). I hope i get the time to look into this soon.

davidmbillie:
Thank you, and I understand. I'd be willing to write more tests for any scenarios that come to find for you as well.

davidmbillie:
@RicoSuter  I'll also add that the conditions I wrote should only affect fringe cases that would have always been failing. Like an IDictionary would have always thrown an exception when doing a hard cast to an IJsonReference. The base functionality of $ref handling should not be touched by these changes.

davidmbillie:
@RicoSuter Sorry to bug, but has this been looked into at all?

luk355:
@RicoSuter, just to add to this, the current version in master does have quite a few issues in ref handling and would really appreciate if this PR can be looked at. Just to name a few:
 * references in `/components/parameters` does not work at all
 * path reference to external file does not work
    ```yaml
       paths:
          "foo":
              $ref: "foo.yaml#/paths/~1foo"
    ```
  * basically any external reference outside of `/components/schemas` is not being resolved

These issues make the tool unusable for any more complicated schemas when the schema is separated into multiple files.

I understand the amount of work you've put into this so far. It would be shame to leave these issues in. Thank you!

and lastly, thanks @davidmbillie for implementing the fix.

RicoSuter:
For path you’d need this, right?
https://github.com/RicoSuter/NSwag/pull/1953

RicoSuter:
Ref: https://github.com/RicoSuter/NJsonSchema/issues/588

davidmbillie:
> For path you’d need this, right?
> [RicoSuter/NSwag#1953](https://github.com/RicoSuter/NSwag/pull/1953)

It looks like so, but I haven't come across any concrete cases with path refs like that yet. If you'd like, I can merge your PR branch into mine and try to write a test for it.

RicoSuter:
This one is in NJS and the other in NSwag so you cannot really merge it, right? :)

RicoSuter:
Ill have some time to look into this tomorrow...

davidmbillie:
> This one is in NJS and the other in NSwag so you cannot really merge it, right? :)

Haha yeah, my bad. I'll still look into writing the unit test for that NJS PR too.

RicoSuter:
Ok, I've updated the code to avoid cloning objects in the setter and also to use the correct serializer... 
Tests still pass and i fixed another issue as well... can you check my changes? Is this ok?

PR #1111, opened 2020-01-07
Expose the ExtensionData's keys as a separate collection

michsco:
The current `ExtensionData` dictionary only allows values to be accessed if the key is known ahead of time. Unfortunately, due to how DotLiquid handles dictionaries, there isn't a way to loop through the keys. This PR will expose the extension data keys for use in templates.


michsco:
@RicoSuter -- you may have more experience with the `Drop` and `IIndexable` interfaces, but creating a new collection was the only way I could figure out to maintain backward compatibility.

RicoSuter:
> you may have more experience with the Drop and IIndexable interfaces, but creating a new collection was the only way I could figure out to maintain backward compatibility

You mean this is the only way to actually iterate the keys in a template?

michsco:
> You mean this is the only way to actually iterate the keys in a template?

Correct.

michsco:
Once you approve this PR, how long does it take to flow back to the NuGet packages?

RicoSuter:
Why did you close/delete the PR?

michsco:
I realized the enumeration wasn't needed for my scenario as I can just as easily collapse my annotations into a single known extension if I am able to get the annotations PR template in.

PR #1119, opened 2020-01-20
Fix issue regarding IAsyncEnumerable treated as JsonObjectType.Object

foriequal0:
It is now treadted as JsonObjectType.Array. This is a temporary
workaround. Eventually the upstream JSON.NET should be fixed to return
JsonArrayContract even if it cannot deserialize.

Fixes: https://github.com/RicoSuter/NSwag/issues/2582

### DIFF
--- a/src/NJsonSchema/Generation/DefaultReflectionService.cs
+++ b/src/NJsonSchema/Generation/DefaultReflectionService.cs
@@ -202,7 +202,8 @@ namespace NJsonSchema.Generation
                 return JsonTypeDescription.CreateForDictionary(contextualType, JsonObjectType.Object, isNullable);
             }
 
-            if (IsArrayType(contextualType) && contract is JsonArrayContract)
+            // TODO: Don't trust JsonArrayContract when contextualType is IAsyncEnumerable<T> until it is fixed
+            if (IsIAsyncEnumerableType(contextualType) || (IsArrayType(contextualType) && contract is JsonArrayContract))
             {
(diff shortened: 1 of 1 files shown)


src/NJsonSchema/Generation/DefaultReflectionService.cs
```diff
@@ -202,7 +202,8 @@ public virtual JsonTypeDescription GetDescription(ContextualType contextualType,
                 return JsonTypeDescription.CreateForDictionary(contextualType, JsonObjectType.Object, isNullable);
             }
 
-            if (IsArrayType(contextualType) && contract is JsonArrayContract)
+            // TODO: Don't trust JsonArrayContract when contextualType is IAsyncEnumerable<T> until it is fixed
+            if (IsIAsyncEnumerableType(contextualType) || IsArrayType(contextualType) && contract is JsonArrayContract)
```
RicoSuter: `if (IsIAsyncEnumerableType(contextualType) || (IsArrayType(contextualType) && contract is JsonArrayContract))`
for safety

foriequal0: Fixed

foriequal0:
Thank you for the quick merge.

RicoSuter:
I'm not sure if this covers all cases but it wont introduce a breaking change so let's release this and see whether it works as expected.

RicoSuter:
Thanks for the pr.

PR #1121, opened 2020-01-21
Allow float to be converted to float instead of double

FelixS90:
This PR fixes [#1122](https://github.com/RicoSuter/NSwag/issues/1122).

If `format` is specified as `float`, the corresponding type in generated C# code will be `float` instead of `double`. I basically just did what @RicoSuter suggested in [this comment](https://github.com/RicoSuter/NSwag/issues/1122#issuecomment-418359039).

Tests in NJsonSchema as well as NSwag executed locally are all passing.

### DIFF
--- a/src/NJsonSchema/Generation/DefaultReflectionService.cs
+++ b/src/NJsonSchema/Generation/DefaultReflectionService.cs
@@ -89,10 +89,14 @@ namespace NJsonSchema.Generation
                 return JsonTypeDescription.Create(contextualType, JsonObjectType.Integer, false, JsonFormatStrings.Long);
             }
 
-            if (type == typeof(double) ||
-                type == typeof(float))
+            if (type == typeof(double))
             {
                 return JsonTypeDescription.Create(contextualType, JsonObjectType.Number, false, JsonFormatStrings.Double);
+            }
+
+            if (type == typeof(float))
+            {
+                return JsonTypeDescription.Create(contextualType, JsonObjectType.Number, false, JsonFormatStrings.Float);
             }
 
             if (type == typeof(decimal))
--- a/src/NJsonSchema.CodeGeneration.CSharp/CSharpTypeResolver.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp/CSharpTypeResolver.cs
@@ -238,6 +238,11 @@ namespace NJsonSchema.CodeGeneration.CSharp
                 return isNullable ? "decimal?" : "decimal";
             }
 
+            if (schema.Format == JsonFormatStrings.Float)
+            {
+                return isNullable ? "float?" : "float";
+            }
+
             return isNullable ? "double?" : "double";
         }
 


src/NJsonSchema/Generation/DefaultReflectionService.cs
```diff
@@ -89,12 +89,16 @@ public virtual JsonTypeDescription GetDescription(ContextualType contextualType,
...
-            if (type == typeof(double) ||
-                type == typeof(float))
+            if (type == typeof(double))
             {
                 return JsonTypeDescription.Create(contextualType, JsonObjectType.Number, false, JsonFormatStrings.Double);
             }
 
+            if (type == typeof(float))
+            {
+                return JsonTypeDescription.Create(contextualType, JsonObjectType.Number, false, JsonFormatStrings.Float);
```
RicoSuter: Is float an officially supported format string of json schema, swagger and openapi?

FelixS90: Thanks for your quick reply.

From what I understand from the JSON Schema w.r.t. [numeric types](https://json-schema.org/understanding-json-schema/reference/numeric.html#number), there is no mandatory `format` specifier at all, i.e. neither for `double` nor for `decimal` or `float`. `string` seems to be the only `type` where `format` is an officially supported field (see https://json-schema.org/understanding-json-schema/reference/string.html#format).

When it comes to OpenAPI, float is an officially supported format as already pointed out correctly in the [original issue post](https://github.com/RicoSuter/NSwag/issues/1122#issue-285422568). From https://swagger.io/specification/#dataTypes:

> Primitive data types in the OAS are based on the types supported by the JSON Schema Specification Wright Draft 00. [...] Primitives have an optional modifier property: format. OAS uses several known formats to define in fine detail the data type being used. [...] Types that are not accompanied by a format property follow the type definition in the JSON Schema. Tools that do not recognize a specific format MAY default back to the type alone, as if the format is not specified.

| type | format | Comments |
| ------------- | ------------- | -------------|
integer | int32 | signed 32 bits |
integer | int64 | signed 64 bits (a.k.a long) |
number | float |  |
number | double | |
integer | int32 | Signed 32-bit integers (commonly used integer type). |
integer | int64 | Signed 64-bit integers (long type). |
|...|...|...|

Independently, I do not see how this affects this part of the code. It has been decided to resolve number formats based on the `format` key specified along with the `type` key. If you do so, this should be consistent, i.e. if `double` is mapped to `double` and `decimal` is mapped to `decimal`, then `float` should be mapped to `float` and not to `double`.

RicoSuter: My comment: 

> Double and float seem to be not defined format types for JSON Schema

so maybe we should change it to 

    if ((settings.SchemaType == SchemaType.Swagger2 || settings.SchemaType == SchemaType.OpenApi3) && type == typeof(float))

RicoSuter: .. so that we don't generate a JSON Schema which is wrong.

FelixS90: So after some more in-depth research:

The [latest draft](https://json-schema.org/specification.html) (2019-09 as of today) states the following in the [document on JSON Schema Validation, Section 7 "A Vocabulary for Semantic Content With "format""](https://tools.ietf.org/html/draft-handrews-json-schema-validation-02#section-7.1):

> Implementations MAY treat "format" as an assertion in addition to an annotation, and attempt to validate the value's conformance to the specified semantics. [...] The value of this keyword is called a format attribute.  It MUST be a string.  A format attribute can generally only validate a given set of instance types.  If the type of the instance to validate is not in this set, validation for this format attribute and instance SHOULD succeed.  All format attributes defined in this section apply to strings, but **a format attribute can be specified to apply to any instance types defined in the data model defined in the core JSON Schema**.

Since `number` is one of the currently six primitive types specified in the [core JSON Schema Instance Data Model](https://tools.ietf.org/html/draft-handrews-json-schema-02#section-4.2.1), it follows that a format attribute can be specified. Since the implementation MAY treat `format` as an assertion, but does not have to, it is subject to the implementation, i.e. NJsonSchema, to decide upon the degree of validation. For more details, refer to the [release notes on the Format Vocabulary](https://json-schema.org/draft/2019-09/release-notes.html#format-vocabulary).

Long story, short: The JSON Schema as generated right now as well as after merging this PR is correct.

Makes sense? What do you think?

RicoSuter: Makes sense to me. Thanks for the investigation.

PR #1123, opened 2020-01-23
Support DataContract private fields and properties

wuzzeb:
I am attempting to update from NSwag 12 to 13 and a bunch of my objects lost most of there fields during the update. I eventually tracked it down to NSwag 13 ignoring private fields declared as DataMember.

In f7f530c54f7b8970104fcba21156d02518885057 a fix was put in to
fix a crash when a property or field did not have MemberInfo.
Unfortunately, this also filtered out private fields and properties
since the check only searched the public fields.  This commit
re-allows private properties and fields by searching the entire
type when looking for a MemberInfo for a property.

### DIFF
--- a/src/NJsonSchema.Tests/Generation/PrivateFieldTests.cs
+++ b/src/NJsonSchema.Tests/Generation/PrivateFieldTests.cs
@@ -1,0 +1,37 @@
+using System.Threading.Tasks;
+using System.Runtime.Serialization;
+using NJsonSchema.Generation;
+using Xunit;
+
+namespace NJsonSchema.Tests.Generation
+{
+  public class PrivateFieldTests
+  {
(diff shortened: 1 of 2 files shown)


src/NJsonSchema.Tests/Generation/PrivateFieldTests.cs
```diff
@@ -0,0 +1,37 @@
+using System.Threading.Tasks;
+using System.Runtime.Serialization;
+using NJsonSchema.Generation;
+using Xunit;
+
+namespace NJsonSchema.Tests.Generation
```
RicoSuter: Use 4 spaces

RicoSuter:
So this is not a breaking change as we always have the lookup by name in the list?

PR #1137, opened 2020-03-03
Typescript: Add setting property EnumStyle with options (Enum, StringLiteral).

trejjam:
This PR adds support for union string literal types instead of enums. It resolves issue https://github.com/RicoSuter/NSwag/issues/1188

### DIFF
--- a/src/NJsonSchema.CodeGeneration.TypeScript/NJsonSchema.CodeGeneration.TypeScript.csproj
+++ b/src/NJsonSchema.CodeGeneration.TypeScript/NJsonSchema.CodeGeneration.TypeScript.csproj
@@ -25,6 +25,7 @@
     <EmbeddedResource Include="Templates\ConvertToClass.liquid" />
     <EmbeddedResource Include="Templates\ConvertToJavaScript.liquid" />
     <EmbeddedResource Include="Templates\Enum.liquid" />
+    <EmbeddedResource Include="Templates\Enum.StringLiteral.liquid" />
     <EmbeddedResource Include="Templates\File.FormatDate.liquid" />
     <EmbeddedResource Include="Templates\File.liquid" />
     <EmbeddedResource Include="Templates\Interface.liquid" />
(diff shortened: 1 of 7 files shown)


src/NJsonSchema.CodeGeneration.TypeScript/TypeScriptTypeResolver.cs
```diff
@@ -149,9 +149,19 @@ private string Resolve(JsonSchema schema, string typeNameHint, bool addInterface
                 if (resolvedType != defaultType)
                 {
                     var keyType = Settings.TypeScriptVersion >= 2.1m ? prefix + resolvedType : defaultType;
-                    if (keyType != defaultType)
+                    if (keyType != defaultType && schema.DictionaryKey.ActualTypeSchema.IsEnumeration)
```
RicoSuter: This looks like a breaking change, no?

trejjam: I do not think so. I tried to find another type where the construct `key in keyof typeof <Identifier>` is useful.
With classes and interfaces it is actually only `key in keyof <Identifier>`. For union types it is only `key in <Identifier>`.

Therefore I do not think it's BC break. I see it more like better safe than sorry.

src/NJsonSchema.CodeGeneration.TypeScript/NJsonSchema.CodeGeneration.TypeScript.csproj
```diff
@@ -25,6 +25,7 @@
     <EmbeddedResource Include="Templates\ConvertToClass.liquid" />
     <EmbeddedResource Include="Templates\ConvertToJavaScript.liquid" />
     <EmbeddedResource Include="Templates\Enum.liquid" />
+    <EmbeddedResource Include="Templates\Enum.StringLiteral.liquid" />
```
RicoSuter: We need to make this an embedded resource

src/NJsonSchema.CodeGeneration.TypeScript/TypeScriptTypeResolver.cs
```diff
@@ -149,9 +149,19 @@ private string Resolve(JsonSchema schema, string typeNameHint, bool addInterface
...
                     {
-                        return $"{{ [key in keyof typeof {keyType}]?: {valueType}; }}";
+                        if (Settings.EnumStyle == TypeScriptEnumStyle.Enum)
```
RicoSuter: Minor thing: I'd also use `else if` as on top TypeScriptGenerator.cs

RicoSuter:
PR looks good to me. Thanks. One minor comment... Is the PR ready after that?

trejjam:
> Is the PR ready after that?

I think so

RicoSuter:
https://github.com/RicoSuter/NJsonSchema/pull/1140

PR #1142, opened 2020-03-08
Add code generation support for C# OneOf schemas

adamjones1:


src/NJsonSchema.CodeGeneration/TypeResolverBase.cs
```diff
@@ -124,7 +124,8 @@ protected virtual bool IsDefinitionTypeSchema(JsonSchema schema)
                    !schema.IsArray &&
                    (schema.IsEnumeration ||
                     schema.Type == JsonObjectType.None ||
-                    schema.Type.HasFlag(JsonObjectType.Object));
+                    schema.Type.HasFlag(JsonObjectType.Object) ||
```
adamjones1: not sure if this might have adverse effects on TypeScript generation?

adamjones1:
Hi @RicoSuter, did you get a chance to look at this one at all?

mladedav:
Hi, this merge request has been approved for a month and a half, would it be possible to get this to a release or at least a pre-release?

adamjones1:
I'm not confident this PR is production-ready yet, eg. see my comment on TypeResolverBase.cs above, I would appreciate some feedback from @RicoSuter first.

jeremyVignelles:
Seems pretty interesting though I didn't have a look in detail. As for your `string | int?` case, I'm not sure how it should be handled.
In [Ooak](https://github.com/jeremyVignelles/ooak), I chose the idea of having the nullable at the top level (i.e. `TypeUnion<string, int>?` instead of `TypeUnion<string, int?>`), because how would you choose between one or the other in the case of `string?|int?` ?

adamjones1:
Yeah it's a difficult problem. It goes broader than nulls too, generally it's the issue of interpreting untagged unions with potentially overlapping types. Eg. `int | long`, `int | int`, and (if from JSON) `string | DateTime`. I'm not sure there's really a good answer. It might be best to limit the scope to schemas which declare a discriminator property and then the unions will be tagged so the issue can be sidestepped.

jeremyVignelles:
My take is that if you have overlapping types, either that's a coding error (in which case just fix the code), or it's intended, and it should deserialize nicely. This is a case I handle in my library:

https://github.com/jeremyVignelles/ooak/blob/7744e6ec70d26d7b8e035126ac895765b12a93fe/Ooak.Testing/PrimitivesTests.cs#L17

The serializer will return a `TypeUnion<TLeft, TRight>.Both` and it's up to the calling code to deal with that. If that's not expected, you could have chosen the `OneOfSerializer` instead of the `AnyOfSerializer`, which would have thrown if both would result in a match.

sylvaneau:
Hi folks,

I've decided to spend some time to work on this issue, i'm currently rebasing the old PR on the head of master, i'm a bit lost with the new inheritence model though...

Do you known which base class `OneOfTemplateModel.cs ` should be based on ? maybe none ?

Pilchard123:
Is there any news on this? I've hit a vendor specification which uses oneOf and none of the client generators I can find work satisfactorily.

* NSwag just doesn't work.
* [OpenAPITools](https://github.com/OpenAPITools/openapi-generator) does work, but requires a lot of dependencies (some of which are old and known to be vulnerable)
* [Swagger Codegen](https://swagger.io/tools/swagger-codegen/) doesn't support post-Framework versions of .NET.

Could inspiration be taken from OpenAPITools's implementation for handling oneOf?

carlreid:
>  none of the client generators I can find work satisfactorily

I have come to the same conclusion. I didn't try [OpenAPITools](https://github.com/OpenAPITools/openapi-generator), as I'd rather not have to install Java just for this.

Think I am just going to have to custom craft the code 😟

PR #1145, opened 2020-03-12
Support for multilingual DisplayAttribute

FarhadGhE:
I needed NJsonSchema in a multilingual project, where each language was added in .resx files. NJsonSchema didn't respect ResourceType argument in DisplayAttribute. So I added support for this scenario. Also it doesn't matter if the resource class has internal or public accessory.
Since Net standard 1.0 doesn't have BindingFlags, I kept the past behavior if target framework is Net standard 1.0

### DIFF
--- a/src/NJsonSchema/NJsonSchema.csproj
+++ b/src/NJsonSchema/NJsonSchema.csproj
@@ -20,6 +20,9 @@
   <PropertyGroup Condition="'$(TargetFramework)' == 'net40'">
     <DefineConstants>LEGACY</DefineConstants>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard1.0'">
+      <DefineConstants>NETSTANDARD10</DefineConstants>
+  </PropertyGroup>
   <PropertyGroup>
     <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
   </PropertyGroup>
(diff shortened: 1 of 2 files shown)


src/NJsonSchema/NJsonSchema.csproj
```diff
@@ -20,6 +20,9 @@
   <PropertyGroup Condition="'$(TargetFramework)' == 'net40'">
     <DefineConstants>LEGACY</DefineConstants>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard1.0'">
```
RicoSuter: I think there is already a const for this (not needed here)

RicoSuter:
Thanks for the PR. Please also add some unit tests.
Doesnt this introduce a breaking change for some people?

wazzamatazz:
I've just run into this exact problem - getting values from resources would be an incredibly useful addition. I don't think it's a breaking change, because the behaviour when you specify `[Display]` but don't specify a `ResourceType` is the same as the current behaviour i.e. use the value of the `Name` property directly instead of looking up a resource.

That said, it would be easier to just let the attribute do the heavy lifting for you, and use the [GetName](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.displayattribute.getname) and [GetDescription](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.displayattribute.getdescription) methods to get the name and description, as these return either the literal value or a resoure depending on how the attribute has been declared. Both are available since .NET Framework 4.0.

RicoSuter:
If there are GetName() and GetDescription() especially for that we should use that and not "reimplement" their logic.

Issue #1133, opened 2020-02-21
Overriden File.liquid template results in no output


smitcham:
Similar to https://github.com/RicoSuter/NJsonSchema/issues/700 I'm attempted to add information to the file template 

```
#if NETSTANDARD2_1
{% template %}
#endif
```
the resulting template is
```
#if NETSTANDARD2_1

#endif
```

if I remove the call to the custom templates, my nswag file generates as I expect it to.

I'm using version 13.2.3 and the netcore31 runtime.

PR #1150, opened 2020-03-19
Remove the exclamation mark to find embedded templates

gillesdb:
Hi @RicoSuter 

Last month I created a PR in NSWag to extend controller templates: https://github.com/RicoSuter/NSwag/pull/2691
But you told me I could use '{% template%}' instead to overwrite a template. 
-> However, this didn't work and you said it was probably a bug.

After some debugging, I found out that the exclamation mark used internally to overwrite templates is not removed when looking for the embedded template.

This pull request removes the exclamation mark at the end of the template name. 

I will also create a pull request for NSwag.
+  With this PR the following issue can also be closed:
https://github.com/RicoSuter/NJsonSchema/issues/1133


RicoSuter:
Is it possible to add a unit test for that?

PR #1153, opened 2020-03-20
Fix UseLeafType with array like objects

trejjam:
I found a bug in previous PR #1148. It does not work well with arrays. Now it should be better


src/NJsonSchema.CodeGeneration.TypeScript/TypeScriptTypeResolver.cs
```diff
@@ -256,9 +258,12 @@ private string ResolveArrayOrTuple(JsonSchema schema, string typeNameHint, bool
                 var isDictionary = schema.Item?.ActualSchema.IsDictionary == true;
 
                 var prefix = addInterfacePrefix && SupportsConstructorConversion(schema.Item) && isObject && !isDictionary ? "I" : "";
-                var itemType = prefix + Resolve(schema.Item, true, typeNameHint);
 
-                return string.Format("{0}[]", GetNullableItemType(schema, itemType)); // TODO: Make typeNameHint singular if possible
+                return string.Join(UnionPipe,
```
RicoSuter: Now you're changing logic in a block where we dont check UseLeafType... is this really correct and does not introduce regressions?

trejjam: We can definitely write it with `UseLeafType` settings. It can harm performance a bit, but without duplicating code.

There can be an issue with nested arrays, but I think they didn't work before

RicoSuter: I think it's save as it only splits in the leaf case.

RicoSuter:
> We can definitely write it with UseLeafType settings

I think it would be better and safer if we'd have that... or not?

Can you create another PR?
I will release NSwag with the current implementation but will update later.

RicoSuter:
Ok, updated. Thanks for the PRs

PR #1165, opened 2020-04-25
Generate enum "StringLiterals" as name implies

onionhammer:
For the given enum

```C#
    public enum AccountType
    {
        User,
        Moderator,
        Administrator
    }
```

Rather than generating

```ts
export type AccountType = 0 | 1 | 2;
```

Which is not very useful, generate 

```ts
export type AccountType =
      "User" 
    | "Moderator" 
    | "Administrator";
```

Note this would be a breaking change to anyone who regenerates their model, but I can't actually see the use for this feature as it exists today.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.TypeScript/Templates/Enum.StringLiteral.liquid
+++ b/src/NJsonSchema.CodeGeneration.TypeScript/Templates/Enum.StringLiteral.liquid
@@ -1,4 +1,6 @@
-﻿{% if HasDescription -%}
+{% if HasDescription -%}
 /** {{ Description }} */
 {% endif -%}
-{% if ExportTypes %}export {% endif %}type {{ Name }} = {% for enumeration in Enums -%}{% if Enums.first.Value != enumeration.Value %} | {% endif %}{{ enumeration.Value }}{% endfor -%};
+{% if ExportTypes %}export {% endif %}type {{ Name }} =
+      {% for enumeration in Enums -%}{% if Enums.first.Value != enumeration.Value %} 
+    | {% endif %}"{{ enumeration.Name }}"{% endfor -%};


RicoSuter:
Thanks for the PR. But I think it is just wrong: The default serialization behavior of JSON.NET is to serialize enums as integer values (instead of string) - this is why we generate 0 | 1 | 2, so that it matches the serializer behavior - you cannot change that in the code generator. If you want strings then you need to add [JsonConverter(typeof(StringEnumConverter))] on the property, enum or globally and then it would also generate the code you expect.

onionhammer:
@RicoSuter In what way is literal `1`, `2`, or `3` a string? Ints are already convertible to enums without any special converter. And in what way is `1` , `2`, or `3` going to tell you something meaningful about an `AccountType`?

Additionally, if, as you propose, you use `StringEnumConverter`, the value you will receive from JSON.NET serialized data for `AccountType.User` is "User", not `0`.. additionally ASPNET will handle "User" as an incoming value and deserialize it just as easily as `0`.

RicoSuter:
If you use StringEnumConverter then the value (and name) are both strings and the AccountType should be with strings, is this not the case?

onionhammer:
@RicoSuter using `StringEnumConverter` has nothing to do with how the swagger json is generated, only with how JSON is serialized from your application. This issue has to do with how the swagger is generated.

Currently the NJsonSchema setting called "StringLiterals" for enum style has **nothing to do** with strings. It generates numbers (0 | 1 | 2)

RicoSuter:
If you dont set StringEnumConverter then the serializer expects ints and you cannot just generate strings because it would not match the serializer settings

onionhammer:
@RicoSuter The json deserializer will work regardless of whether you set StringEnumConverter, it's smart enough to handle int or string.

RicoSuter:
> The json deserializer in asp.net will work regardless of whether you set StringEnumConverter, it's smart enough to handle int or string.

Ok that might be true for asp.net but if the spec specifies the values as 1,2,3 then only these values are allowed and we cannot just generate strings.. you cannot assume that it works with the names (x-enum-names) with any backend.

onionhammer:
@RicoSuter Can you give me an example use-case of when this would come in handy

```
interface Model {
  SomeEnum: 0 | 1 | 2
}
```

What does 0 mean? What does 1 mean? What does 2 mean? From a typescript POV they are just meaningless numbers.

Additionally, can you please explain what `0 | 1 | 2` has to do with "StringLiterals" ?

onionhammer:
@RicoSuter Looks like you fixed this in a separate PR, not sure what the disagreement was about though; the latest version of NJsonSchema now generates string literals properly.

PR #1169, opened 2020-05-06
Fix ToSampleJson issues for Array

stg609:
Current ToSampleJson will throw an null reference exception when invoke the ToSampleJson on an  array schema like below:
`{
  "type": "array",
  "items": [
    {
      "type": "number"
    },
    {
      "type": "string"
    },
    {
      "type": "string",
      "enum": ["Street", "Avenue", "Boulevard"]
    },
    {
      "type": "string",
      "enum": ["NW", "NE", "SW", "SE"]
    }
  ]
}`

Besides, ToSampleJson will also insert a null item for array.

This PR just fix above issues.


RicoSuter:
Thanks for the PR. This area (sample data generator) still requires quite some improvements and would also be nice to add unit tests..

PR #1170, opened 2020-05-06
Patch/improved enum handling

truck0321:
Improved enum handling resolves RicoSuter/NJsonSchema#1357

### DIFF
--- a/src/NJsonSchema.CodeGeneration/DefaultEnumNameGenerator.cs
+++ b/src/NJsonSchema.CodeGeneration/DefaultEnumNameGenerator.cs
@@ -33,36 +33,36 @@ namespace NJsonSchema.CodeGeneration
             switch (name)
             {
                 case ("="):
-                    name = "eq";
+                    name = "Eq";
                     break;
                 case ("!="):
-                    name = "ne";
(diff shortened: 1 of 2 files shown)


truck0321:
This PR resolves RicoSuter/NSwag#1357

src/NJsonSchema.CodeGeneration/DefaultEnumNameGenerator.cs
```diff
@@ -24,11 +24,47 @@ public class DefaultEnumNameGenerator : IEnumNameGenerator
...
+
             if (string.IsNullOrEmpty(name))
             {
                 return "Empty";
             }
 
+            switch (name)
+            {
+                case ("="):
+                    name = "eq";
```
RicoSuter, line 36: I think we should make this upper case so that "-Foo" gets converted to "MinusFoo" => the default output should be Upper Camel Case, please add some tests so showcase the change => there should already be some tests somewhere in the project.

truck0321: Added test coverage and changed output to Camel Case.

Issue #1181, opened 2020-05-22
Index out of Range on Convert to Camel Case methods


caleblanchard:
Receiving an Index out of range error when passing an input string with a length of 1 into `ConversionUtilities.ConvertToLowerCamelCase()` and `ConversionUtilities.ConvertToUpperCamelCase()`.

Submitting a PR shortly to resolve the issue.

PR #1182, opened 2020-05-22
Fix index out of range error on convert to camel case methods

caleblanchard:
When passing a string of length 1 to the camel case methods, it was throwing an index out of range exception. This PR is to fix that issue: #1181


RicoSuter:
Thanks for the PR!

PR #1190, opened 2020-06-10
Ensure nullable marker isn't on enum default value

jeffbdye:
Hey, thanks for maintaining such a great repo + toolset here! 

I'm hoping to address #1141. I'm not entirely certain this is the best way to approach fixing what's happening here, effectively this condition was previously overriding the intentionally set `isNullable = false` in `ValueGeneratorBase.GetEnumDefaultValue`. There doesn't seem to be a ton of testing around `GenerateOptionalPropertiesAsNullable` so I'm not particularly confident in this change, but I can try to add more to establish expected behavior here. 

Optimally, it'd be nice if the type resolution knew if it was being used to resolve a type for a property type or for the value of a property - is there anything on the `JsonSchema` classes that'd indicate that? If not, another possible way to make this happen might be to make `CSharpValueGenerator.GetEnumDefaultValue` more explicit in removing possible nullability (though I don't want to encroach too much on what seems to be the responsibility of the TypeResolver classes).

Please let me know how y'all would like to proceed, I was mainly hoping to call attention to this issue since it came up recently for us.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/CSharpTypeResolver.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp/CSharpTypeResolver.cs
@@ -70,7 +70,8 @@ namespace NJsonSchema.CodeGeneration.CSharp
             // Primitive schemas (no new type)
             if (Settings.GenerateOptionalPropertiesAsNullable &&
                 schema is JsonSchemaProperty property &&
-                !property.IsRequired)
+                !property.IsRequired &&
+                !schema.IsEnumeration)
             {
                 isNullable = true;
             }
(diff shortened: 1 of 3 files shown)


src/NJsonSchema.CodeGeneration.CSharp/CSharpTypeResolver.cs
```diff
@@ -70,7 +70,8 @@ public string Resolve(JsonSchema schema, bool isNullable, string typeNameHint, b
             // Primitive schemas (no new type)
             if (Settings.GenerateOptionalPropertiesAsNullable &&
                 schema is JsonSchemaProperty property &&
-                !property.IsRequired)
+                !property.IsRequired &&
+                !schema.IsEnumeration)
```
RicoSuter, line 74: This doesnt look right, because an enum can also be nullable - or am I missing something?

RicoSuter:
Fixed with a small "hack".

jeffbdye:
Right on, thanks!

RicoSuter:
Released in njs and nswag v13.6.1

PR #1201, opened 2020-06-18
Improved generated code with GenerateNullableReferenceTypes

### LINKED ISSUE #1195: Null Reference Types cause CS8669
```
warning CS8669: The annotation for nullable reference types
    should only be used in code within a '#nullable' annotations context. Auto-generated code
    requires an explicit '#nullable' directive in source.
```
Please add explicit #nullable when null reference types are used.  This happens even with Nullable is turned on at the project level.

jeremyVignelles:
Added a `#nullable enable` directive in generated C# source only if `GenerateNullableReferenceTypes` is set to `true`.

Forced properties without a default value to `default(T)!`. This does not change the existing behavior, but there is no warning if the user forgets to set the property.

Fixes #1195

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/Models/FileTemplateModel.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp/Models/FileTemplateModel.cs
@@ -14,6 +14,9 @@ namespace NJsonSchema.CodeGeneration.CSharp.Models
         /// <summary>Gets or sets the namespace.</summary>
         public string Namespace { get; set; }
 
+        /// <summary>Gets or sets a value indicating whether the C#8 nullable reference types are enabled for this file.</summary>
+        public bool GenerateNullReferenceTypes { get; set; }
+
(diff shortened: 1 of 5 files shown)


src/NJsonSchema.CodeGeneration.CSharp/Models/FileTemplateModel.cs
```diff
@@ -14,6 +14,9 @@ public class FileTemplateModel
         /// <summary>Gets or sets the namespace.</summary>
         public string Namespace { get; set; }
 
+        /// <summary>Gets or sets a value indicating whether the C#8 nullable reference types are enabled for this file.</summary>
+        public bool GenerateNullReferenceTypes { get; set; }
```
jeremyVignelles, line 18: oops, should be `GenerateNullableReferenceTypes`

RicoSuter:
I hope to have time tomorrow to look into this..

Issue #1031, opened 2019-07-17
System.ObsoleteAttribute on properties


pschwarzpp:
We would like to prepare users of our API to a change regarding one model.

AFAICT, OpenAPI v3 added support for deprecation on property level (https://github.com/OAI/OpenAPI-Specification/issues/701, https://github.com/OAI/OpenAPI-Specification/issues/524).

As it seems, currently neither `aspNetCoreToOpenApi` nor `openApiToCSharpClient` support deprecation of a single property in a model (the generated spec / code is not affected by the attribute).

Should setting `ignoreObsoleteProperties` to true cause the property to be omitted from the generated spec?

RicoSuter:
Should search in the repo for ObsoleteAttribute

pschwarzpp:
I'm sorry for the missing information regarding my last sentence (`ignoreObsoleteProperties`).

This would help us (but general support for depreciation would be preferable), but in my simple tests it did not work.

For the test I used NSwag.Sample.NETCore21. Changing `ignoreObsoleteProperties` to `true` and adding `[Obsolete]` to `NSwag.Sample.Pet.HasVaccinations` has no effect on the generated swagger spec.

Nswag used:
NSwag command line tool for .NET Core NetCore21, toolchain v13.0.4.0 (NJsonSchema v10.0.21.0 (Newtonsoft.Json v11.0.0.0)) (installed from npm)

PR #1207, opened 2020-06-24
Deprecation improvements

ackginger:
This hopefully addresses the crux of this issue:
https://github.com/RicoSuter/NJsonSchema/issues/1031

and will unblock this enhancement in NSwag:
https://github.com/RicoSuter/NSwag/issues/2595

Along with a change to use the native deprecated attribute in OpenAPI 3.0 instead of the x-deprecated extension (#1208)

Ultimately this allows the retention of `[Obsolete("Reason")]` on classes and properties between code and generated client.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/Templates/Class.liquid
+++ b/src/NJsonSchema.CodeGeneration.CSharp/Templates/Class.liquid
@@ -13,6 +13,9 @@
 {% if InheritsExceptionSchema -%}
 [Newtonsoft.Json.JsonObjectAttribute]
 {% endif -%}
+{% if IsDeprecated -%}
+[System.Obsolete{% if HasDeprecatedMessage %}("{{ DeprecatedMessage }}"){% endif %}]
+{% endif -%}
 {% template Class.Annotations %}
 {{ TypeAccessModifier }} {% if IsAbstract %}abstract {% endif %}partial class {{ClassName}} {% template Class.Inheritance %}
 {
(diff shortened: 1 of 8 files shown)


RicoSuter:
Lgtm 👍 thx

src/NJsonSchema/JsonSchema.cs
```diff
@@ -441,10 +441,14 @@ public JsonObjectType Type
         [JsonProperty("minProperties", DefaultValueHandling = DefaultValueHandling.IgnoreAndPopulate)]
         public int MinProperties { get; set; }
 
-        /// <summary>Gets or sets a value indicating whether the schema is deprecated (Swagger and Open API only).</summary>
+        /// <summary>Gets or sets a value indicating whether the schema is deprecated (native in Open API 'deprecated', custom in Swagger 'x-deprecated').</summary>
```
RicoSuter, line 444: Change to "custom in Swagger/JSON Schema"

src/NJsonSchema.CodeGeneration.CSharp/Templates/Class.liquid
```diff
@@ -13,6 +13,9 @@
 {% if InheritsExceptionSchema -%}
 [Newtonsoft.Json.JsonObjectAttribute]
 {% endif -%}
+{% if IsDeprecated -%}
+[System.Obsolete{% if HasDeprecatedMessage %}("{{ DeprecatedMessage }}"){% endif %}]
```
RicoSuter, line 17: Here we need to escape the message - it might contains special chars which break the compiler. 
We can use this code here: 

image

Simple way is to convert in the model class, the right solution would be to add a transformer like `| csharpdocs` and use it like this in the liquid tpl: 

image

image

e.g. `| literal`
e

src/NJsonSchema/JsonSchema.cs
ackginger: Done.

src/NJsonSchema.CodeGeneration.CSharp/Templates/Class.liquid
ackginger: Added the transformer as I assume it might be useful in future.

PR #1210, opened 2020-06-25
Make code generation respectful to null values

asvishnyakov:


asvishnyakov:
@RicoSuter please review

PR #1211, opened 2020-06-25
* Changed IsPropertyIgnored to protected virtual

rbvictor:
It is useful for us to override this method and customize the criteria for ignoring properties.


RicoSuter:
Build failing, probably because xml docs are missing

Issue #1203, opened 2020-06-20
NJsonSchema.Yaml references deprecated package YamlDotNet.Signed 


MatthiasLaroche:
The nuget package YamlDotNet.Signed is marked as deprecated since version 6.0.0, and all versions including 5.1.0 are now unlisted.

https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema.Yaml/NJsonSchema.Yaml.csproj#L22

The YamlDotNet package is at version 8.1.2 currently, and the package has been signed since version 6.0.0 (15/03/2019).

Currently, any developer using the non-deprecated YamlDotNet version 6.0.0 - 8.1.2 will have conflicting package versions.

PR #1214, opened 2020-06-26
Patches/upgrade obsolete yamldotnet

MatthiasLaroche:
Hello,

I wanted to change the obsolete package of YamlDotNet.Signed 5.1.0 to the latest YamlDotNet 8.1.2 to avoid conflicts in my dependencies.
https://github.com/RicoSuter/NJsonSchema/issues/1203

While doing so, I saw the net45 tests were not running because the xunit runner is no longer compatible with net45 and requires net452. So I fixed that and saw that an old net45 test was using the CodeDomProvider which is not compatible with C# 6+ and cannot understand the read-only properties generated by NJsonSchema. So I used the latest version of Roslyn (1.3.2) compatible with net452 for this test instead, copying the netcoreapp2.0 test.
Now every tests seems to run smoothly again.


RicoSuter:
Thx, lets hope we dont break someone...

PR #1231, opened 2020-08-18
DefaultTypeNameGenerator remove illegal characters

archigo:
Replace illegal characters with an underscore, such that there are no names generated that cannot compile.


archigo:
It is potentially breakking. Unicode characters up to unicode3.0 are valid in class names, but i didn't find any good way to check for only unicode 3.0

PR should also be changed to apply the same logic on method and property names.

RicoSuter:
To me it looks ok, as you should only have [a-zA-Z0-9]* anyway.. 
Should we merge now or would you also update the CSharpPropertyNameGenerator ?

archigo:
Let's merge it. If it proves necessary in the future to also clean property names, it should be easy to move the "RemoveIllegalCharacters" method out as a shared and use it from both places.

PR #1236, opened 2020-08-25
add function for createInstance

itvius:
Hello, i have a problem with createInstance. In your variant, if we have more 5000 data item, createInstance will work more 7 seconds. For fix this problem create function createInstance1. Thanks, and sorry for bad language.


RicoSuter:
Can you replace the function so that we can see the diff? Maybe also explain why this one is better/faster and what breaking change we can expect (or is it the exact same behavior?)?

itvius:
Change create instance. Is it the exact same behavior, but faster.

PR #1237, opened 2020-08-26
Ignore case for discriminator

Pzixel:
Currently Json.Net by default allows deserialize both `"foo":1` and `"Foo":1` into `class Root { public int Foo {get;set;} }`, but surprisingly JsonInheritanceConverter - doesn't.

Maybe it worth to add a constructor argument for specifying which StringComparison should it use but I think ignore case by default could be enough.

### DIFF
--- a/src/NJsonSchema/Converters/JsonInheritanceConverter.cs
+++ b/src/NJsonSchema/Converters/JsonInheritanceConverter.cs
@@ -163,7 +163,7 @@ namespace NJsonSchema.Converters
                 return null;
             }
 
-            var discriminator = jObject.GetValue(_discriminator).Value<string>();
+            var discriminator = jObject.GetValue(this._discriminator, StringComparison.OrdinalIgnoreCase).Value<string>();
             var subtype = GetDiscriminatorType(jObject, objectType, discriminator);
 
             var objectContract = serializer.ContractResolver.ResolveContract(subtype) as JsonObjectContract;


RicoSuter:
I dont think the OpenAPI spec allows case insensitive discriminators...

Pzixel:
This is how Json.Net works by default. So different case actually work when real JSON get (de)serialized.

Being said, if it goes against the spec (which i'm not sure it does) then it could be at least configurable with spec-compatible setting by default.

mirinkinen:
I also have this problem, or a related problem at least. 

I have all sorts of JSON in database that needs to be deserialized with custom discriminator fields and values. The same data needs to serialized for REST API clients, BUT the discriminator fields needs to be in different form than in the database. For example, I might have discriminator field of "Type" with value "text", which is first deserialized and then serialized again, but in this case, but the discriminator field name should be "type", all lower-case. So basically, I need different serialized forms for the same object, where the only difference is the casing of the discriminator field name.

This is not a huge problem of the API, but it looks weird when serialized objects have one field with capital letters, for example:

`{
  "id": 123,
  "age": "25",
  "Type": "person"
}`

I think the only way to do this, is to implement the whole JsonConverter abstract class, which I did. So I deserialized the discriminator like this:

`var discriminator = jObject.GetValue(_discriminator, StringComparison.InvariantCultureIgnoreCase).Value<string>();`

and serialized it for the REST API like this:

`jObject[_discriminator.ToLower()] = JToken.FromObject(GetDiscriminatorValue(value.GetType()));`

Not sure if this also could be tackled with `Newtonsoft.Json.JsonSerializer` settings or custom implementation of it.

Pzixel:
Would you like to merge this if I add a configurable property (with default as it is now)?

mirinkinen:
I was wrong. If the class is not inherited from `JsonInheritanceConverter`, then the generated OpenAPI documentation will not show any discriminator mapping. It is mandatory to inherit from `JsonInheritanceConverter` and NOT from `JsonConverter`, if custom discriminator implementation is needed.

RicoSuter:
I'd say that Newtonsoft's behavior is just wrong, but as this is a Newtonsoft converter I'm ok with this change. 
If there is an exact match it will win anyway so i dont see a problem with that...

The exact
        //     property name will be searched for first and if no matching property is found
        //     then the System.StringComparison will be used to match a property.

PR #1255, opened 2020-09-26
Fix _readTypeProperty jObject.GetValue(_discriminator) returned null.

FreeFrags:
Fix _readTypeProperty jObject.GetValue(_discriminator) returned null.


RicoSuter:
Do we also need to fix this in the generated converter?

https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration.CSharp/Templates/JsonInheritanceConverter.liquid#L77``

PR #1258, opened 2020-09-29
Fix reading of external refs from paths with a # character in them

### LINKED ISSUE #1257: Problem loading a file with # on its paths
Whenever there is a `$ref` which resolves to a full path with a `#` character in it, the library will fail to load it (as it treats it as fragment instead of file path). The `$ref` itself doesn't need to (and shouldn't) have a `#` character in it (as part of the path, which would be incorrect URI as it should be escaped as `%23`), if the relative path in the `$ref` resolves to a path with `#` in it, the problem will surface.

### Solution

The source of the problem (sort-of see below) is how the `JsonReferenceResolver.ResolveFileReferenceWithAlreadyResolvedCheckAsync` splits the `fullJsonPath`:

```cs
var arr = Regex.Split(fullJsonPath, @"(?=#)");
```

Changing it to:

```cs
string[] arr;
var i = fullJsonPath.LastIndexOf('#');
if (i == -1)
{
    arr = new[] { fullJsonPath };
}
else
{
    arr = new[] { fullJsonPath.Substring(0, i), fullJsonPath.Substring(i) };
}
```

Resolves the issue in my case. But I'm not sure if it is universal enough to not cause any other issues (but according to JSON Pointer spec, `#` should only appear as the first character and having it somewhere else is an ill-formatted JSON Pointer).

Another solution would be to change how paths are resolved internally and use an `Uri` class (which would escape the `#`).

shadow-cs:
Fixes #1257


src/NJsonSchema.Tests/References/LocalReferencesTests.cs
```diff
@@ -123,9 +123,51 @@ public async Task When_reference_is_registered_in_custom_resolver_it_should_not_
             Assert.NotNull(schema);
         }
 
+        [Theory]
+        [InlineData("b%23r", Skip = "Not working ATM")] // Escaped well-formed JSON Pointer
```
shadow-cs, line 127: I did find this is currently not working, I guess I can fix as well but I this is kind-of different issue ;-)

RicoSuter: Should this even be supported?

RicoSuter: Let's fix this in another PR.

RicoSuter:
I updated your code a bit to align it with the changes of the other PR.

RicoSuter:
Thanks for the PR.

src/NJsonSchema.Tests/References/LocalReferencesTests.cs
shadow-cs: Should I try to fix it?

RicoSuter: Should this be supported?

shadow-cs: I guess so...

This is the correct way special characters should be escaped in JSON Pointer and the same applies for JSON Schema/Swagger no?

As https://tools.ietf.org/html/rfc6901#section-6 states:

> while percent-encoding those characters not allowed by the fragment

I did test other validators and they support both percent encoding and special characters (like `#`) in the JSON Pointer part of URI (even though this is not exactly according to spec). In other words if you want to be as compatible as possible, supporting this is a way to go.

RicoSuter: Ok then we probably should support that and do the split on # and the other sequence everywhere. Do you want to create a PR?

shadow-cs: We're still talking about `src/NJsonSchema/JsonReferenceResolver.cs` right? If so then sure, I can make the adjustments...

RicoSuter: Yep exactly

shadow-cs: Ok, will do

PR #1262, opened 2020-10-02
Support Array default

xobust:
When using ToSampleJson() the arrays will be populated with
the default value of the item.
```
{
  "title": "Foo",
  "type": "object",
  "properties": {
    "arr": {
      "type": [
        "array",
        "null"
      ],
      "default": [1,2,3],
      "items": {
        "type": "integer",
        "format": "int32"
      }
    }
  }
}
```
Results in:
`{{
  "arr": [
    0
  ]
}}`

This PR will instead populate it by the default if it exists. 
`
{{
  "arr": [
    1,2,3
  ]
}}
`


RicoSuter:
Looks good. Thanks for the PR.

Issue #1273, opened 2020-11-05
Make async methods cancellable (via cancellationToken)


Bosch-Eli-Black:
Hello 🙂

Could you please add an overload to` JsonSchema::FromFileAsync that takes a `cancellationToken`, so that this method can be gracefully cancelled?

Thanks 🙂

RicoSuter:
Good point, also warned here :-)

https://apimundo.com/organizations/nuget-org/nuget-feeds/public/packages/NJsonSchema/versions/latest?tab=recommendations&type=NJsonSchema.JsonSchema

PR #1277, opened 2020-11-10
support cancellation token for async methods without breaking change

ChangeTheCode:
- add Cancellation token support with default cancellation token to do not have breaking change 
- Fix warnings inside of unit test asserts 
- remove / for comments to fix warnings
- tag methods without cancellation token as obsolete  
#1273

### DIFF
--- a/src/NJsonSchema/JsonSchema.cs
+++ b/src/NJsonSchema/JsonSchema.cs
@@ -12,6 +12,7 @@ using System.Collections.ObjectModel;
 using System.Linq;
 using System.Reflection;
 using System.Text.RegularExpressions;
+using System.Threading;
 using System.Threading.Tasks;
 using Namotion.Reflection;
 using Newtonsoft.Json;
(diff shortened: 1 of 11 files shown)


src/NJsonSchema/JsonSchema.cs
```diff
@@ -134,71 +135,79 @@ public static JsonSchema FromSampleJson(string data)
...
         /// <param name="referenceResolverFactory">The JSON reference resolver factory.</param>
+        /// <param name="cancellationToken">The cancellation token</param>
         /// <returns>The JSON Schema.</returns>
         /// <exception cref="NotSupportedException">The HttpClient.GetAsync API is not available on this platform.</exception>
-        public static async Task<JsonSchema> FromUrlAsync(string url, Func<JsonSchema, JsonReferenceResolver> referenceResolverFactory)
+        public static async Task<JsonSchema> FromUrlAsync(string url, Func<JsonSchema, JsonReferenceResolver> referenceResolverFactory, CancellationToken cancellationToken = default)
         {
-            var data = await DynamicApis.HttpGetAsync(url).ConfigureAwait(false);
-            return await FromJsonAsync(data, url, referenceResolverFactory).ConfigureAwait(false);
+            var data = await DynamicApis.HttpGetAsync(url).ConfigureAwait(false);
```
RicoSuter, line 177: I think it's important to add the ct in the DynamicApis as this is where the I/O bound stuff actually is...

RicoSuter: Added here: https://github.com/RicoSuter/NJsonSchema/pull/1277

PR #1287, opened 2020-11-26
fix(NJsonSchema.CodeGeneration.TypeScript): add explicit '| null' in createInstance return type

rars:
When compiling the generated TypeScript code with strict null checking enabled, the return type of `createInstance` is highlighted as an error as its return type is `T` even though `null` can be returned.

This PR adds `| null` to the return type.


RicoSuter:
Thanks for the PR.

PR #1289, opened 2020-11-30
Use nullable reference for AnyType if required

jacobpovar:
Hi. When using C# code generation from openapi3.0 format and given the following :

- GenerateNullableAnnotations option is enabled, 
- and type of nullable object isn't specified:
```
propertyName:
  type: object
  nullable: true
```

then Nswag generates code with `object` type for propertyName, not `object?`.

Changing AnyType to `object?` in code generation settings doesn't help either since it results with following code:
```
public object? PropertyName {get;set;} = new object?();
```
which can't be compiled.

This PR changes type resolver behavior by append `?` to AnyType.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp.Tests/NullableReferenceTypesTests.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp.Tests/NullableReferenceTypesTests.cs
@@ -1,0 +1,152 @@
+﻿using System.ComponentModel.DataAnnotations;
+using System.Threading.Tasks;
+using NJsonSchema.CodeGeneration.CSharp;
+using NJsonSchema.Generation;
+using Xunit;
+
+namespace NJsonSchema.CodeGeneration.Tests.CSharp
+{
+    public class NullableReferenceTypesTests
+    {
+        private class ClassWithRequiredObject
+        {
+            public object Property { get; set; }
+
+            [Required]
+            [Newtonsoft.Json.JsonProperty("property2", Required = Newtonsoft.Json.Required.Always)]
+            public object Property2 { get; set; }
+        }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


RicoSuter:
@jeremyVignelles can you do a review?
It looks good to me.

RicoSuter:
@jacobpovar Can you also add some tests covering this?

jeremyVignelles:
Does the code still generate `= new object();` ?

I don't think it's the right thing to return(should be null), but that would likely break things to change it now.

If there is `= new object();`, can it still be null or will it just be skipped by deserialization?
If it can't be null, there's no point of making the type nullable

jacobpovar:
@RicoSuter @jeremyVignelles thank you for reviewing this request

@RicoSuter sure, added couple of tests.

@jeremyVignelles 

An object will have `new object()` initializer only if it's required.
when reading jsonschema from C# class then both optional and required fields will have [`= default!` initializer](https://github.com/RicoSuter/NJsonSchema/blob/5f129d99e87111a6b4430db7617002da0ee75608/src/NJsonSchema.CodeGeneration.CSharp.Tests/NullableReferenceTypesTests.cs#L64).

when reading schema from json (the difference is that properties have explicit object type now) an optional field still has `= default!` initializer and will remain null. 
[a required field will have `new object()` initializer as it had before my changes](https://github.com/RicoSuter/NJsonSchema/blob/5f129d99e87111a6b4430db7617002da0ee75608/src/NJsonSchema.CodeGeneration.CSharp.Tests/NullableReferenceTypesTests.cs#L149).

So, if I'm not mistaken, If property is not set, it will remain null after deserialization. If it's present, then it won't be null.

jeremyVignelles:
I didn't check but that sounds good 🙂 Thanks for the explanations

PR #1297, opened 2020-12-18
fix(generator): fix round issue for sample date generator

ChangeTheCode:
- generator for sample data cut decimal points to 1 
- now generates also decimal numbers with e.g. 5 numbers after the point


RicoSuter:
Thanks for the PR.

PR #1299, opened 2020-12-27
Array of nullable items are now automatically detected

jeremyVignelles:
There is no need to use [ItemsCanBeNull] in an NRT context.
Fixes https://github.com/RicoSuter/Namotion.Reflection/issues/36


src/NJsonSchema/Generation/JsonSchemaGenerator.cs
```diff
@@ -580,7 +580,7 @@ protected virtual string[] GetTypeProperties(Type type)
             var itemType = jsonSchemaAttribute?.ArrayItem ?? contextualType.OriginalType.GetEnumerableItemType();
             if (itemType != null)
             {
-                var contextualItemType = itemType.ToContextualType();
+                var contextualItemType = contextualType.ElementType ?? itemType.ToContextualType();
```
jeremyVignelles, line 583: not sure why `contextualType.ElementType` is `null` for some other cases. Maybe there's something that should be fixed in Namotion.Reflection ?

For example, this doesn't work with `List<string?>`

jeremyVignelles:
Not sure why the CI failed, works on my machine...

src/NJsonSchema/Generation/JsonSchemaGenerator.cs
RicoSuter: Here we need to use the new EnumerableItemType, right?

src/NJsonSchema.Tests/Generation/ArrayGenerationTests.cs
```diff
@@ -29,6 +29,27 @@ public async Task When_property_is_JArray_then_schema_with_any_array_is_generate
             Assert.True(arrayProperty.Item.ActualTypeSchema.IsAnyType);
         }
 
+#nullable enable
+        public class ClassWithArrayOfNullable
+        {
+            public string?[] Array { get; set; } = new string?[0];
```
jeremyVignelles, line 35: Can we add a test for `List<string?>` now that it's supported by Namotion.Reflection ?

RicoSuter: Added. Build fails in CI but local it's fine... investigating...

RicoSuter:
I have the feeling it's because we are using 2.2 SDK and maybe this stuff produces different/better/correct NRT attributes with 5.0 SDK

Issue #1013, opened 2019-06-15
Use System.Text.Json in generated C# models (DTOs)


RicoSuter:

RicoSuter:
Ref 
https://github.com/RicoSuter/NJsonSchema/pull/1333
https://github.com/RicoSuter/NJsonSchema/pull/1331

xperiandri:
That would be super helpful

RicoSuter:
See https://github.com/RicoSuter/NSwag/issues/2243#issuecomment-780438323

PR #1308, opened 2021-01-26
Add System.Text.Json support in DTO generation

RicoSuter:
First iteration for #1013


stijnherreman:
@RicoSuter is `SystemTextJson.JsonInheritanceConverter` meant to live in the test project?
Asking because I'd like to use it somewhere in a project that already uses NSwag/NJsonSchema anyway.

PR #1312, opened 2021-02-08
Add System.Text.Json ExtensionData attribute tests

deceptiveSimplicity:
As requested in https://github.com/RicoSuter/NSwag/issues/3280


RicoSuter:
Updated and fixed test. Please review.

RicoSuter:
Decided to merge :-)
Ping me if there is a problem with the PR.

deceptiveSimplicity:
Sorry for being late...
The change looks good to me. :)
I haven't tested it with NSwag and the model in https://github.com/RicoSuter/NSwag/issues/3280, but I think this change should fix the problem. I'll give more feedback in the issue when I tested it.

RicoSuter:
NSwag is not yet updated with this change... will update and release probably today

deceptiveSimplicity:
I know, that's why I haven't tested it yet. It is a bit more complicated to test NSwag and NJsonSchema from source in my solution. I will wait for your release. Thank you very much for your work.

PR #1313, opened 2021-02-12
Use DisplayAttribute methods when getting name and description

wazzamatazz:
Related to #1145.

`DisplayAttribute` has [GetName](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.displayattribute.getname) and [GetDescription](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.displayattribute.getdescription) methods that will return the `Name` and `Description` property values respectively if no `ResourceType` has been specified on the attribute, or will look up resource values if a `ResourceType` has been specified.

This PR updates schema generation to use the above methods when a `DisplayAttribute` is used, which allows localised resources to be used when generating schemas in addition to hard-coded name and description values.


src/NJsonSchema/Generation/JsonSchemaGenerator.cs
```diff
@@ -351,9 +351,10 @@ public virtual void ApplyDataAnnotations(JsonSchema schema, JsonTypeDescription
             var contextualType = typeDescription.ContextualType;
 
             dynamic displayAttribute = contextualType.ContextAttributes.FirstAssignableToTypeNameOrDefault("System.ComponentModel.DataAnnotations.DisplayAttribute");
-            if (displayAttribute != null && displayAttribute.Name != null)
+            if (displayAttribute != null)
             {
-                schema.Title = displayAttribute.Name;
+                // GetName returns null if the Name property on the attribute is not specified.
+                schema.Title = displayAttribute.GetName();
```
RicoSuter, line 357: Is using GetName() a breaking change from what we had before?
Maybe we should add a null check here as well (as before) as title might already be set and gets resetted to null otherwise.

src/NJsonSchema/Infrastructure/TypeExtensions.cs
```diff
@@ -73,9 +73,14 @@ public static string GetDescription(this CachedType type, DescriptionAttributeTy
             else
             {
                 dynamic displayAttribute = attributes.FirstAssignableToTypeNameOrDefault("System.ComponentModel.DataAnnotations.DisplayAttribute");
-                if (displayAttribute != null && !string.IsNullOrEmpty(displayAttribute.Description))
+                if (displayAttribute != null)
                 {
-                    return displayAttribute.Description;
+                    // GetDescription returns null if the Description property on the attribute is not specified.
+                    var description = displayAttribute.GetDescription();
```
RicoSuter, line 79: Is using GetDescription() a breaking change from what we had before?

src/NJsonSchema/Generation/JsonSchemaGenerator.cs
wazzamatazz: GetName() will return the value of the Name field if the ResourceType property on the DisplayAttribute is null, so in this circumstance, the behaviour is the same.

If a ResourceType property is specified on the DisplayAttribute, then GetName() will retrieve the value from the ResourceType instead e.g.

```csharp
public class MyClass {
  // No ResourceType specified: DisplayAttribute.GetName() returns "Some_Property"
  [Display(Name = "Some_Property")]
  public string SomeProperty { get; set; }
}


public class MyOtherClass {
  // ResourceType specified: DisplayAttribute.GetName() returns MyResources.Some_Property
  [Display(ResourceType = typeof(MyResources), Name = "Some_Property")]
  public string SomeProperty { get; set; }
}
```

Current behaviour does not have a check to see if the title has already been set.

src/NJsonSchema/Infrastructure/TypeExtensions.cs
wazzamatazz: GetDescription() will return the value of the Description field if the ResourceType property on the DisplayAttribute is null, so in this circumstance, the behaviour is the same.

If a ResourceType property is specified on the DisplayAttribute, then GetDescription() will retrieve the value from the ResourceType instead which is the intended behaviour of this change e.g.

```csharp
public class MyClass {
  // No ResourceType specified: DisplayAttribute.GetDescription() returns "Some_Property_Desc"
  [Display(Description = "Some_Property_Desc")]
  public string SomeProperty { get; set; }
}


public class MyOtherClass {
  // ResourceType specified: DisplayAttribute.GetDescription() returns MyResources.Some_Property_Desc
  [Display(ResourceType = typeof(MyResources), Name = "Some_Property_Desc")]
  public string SomeProperty { get; set; }
}
```

src/NJsonSchema/Generation/JsonSchemaGenerator.cs
wazzamatazz: Committed a check to ensure that the `Name` property is not null before calling GetName()

src/NJsonSchema/Infrastructure/TypeExtensions.cs
wazzamatazz: Committed a check to ensure that the `Description` property is not null before calling GetDescription()

src/NJsonSchema/Generation/JsonSchemaGenerator.cs
RicoSuter: Instead of a .Name check and know the internals of GetName() I think it would be better to do: 

```
if (displayAttribute != null)
{
     var name = displayAttribute.GetName();
     if (name != null)
     {
         schema.Title = name;
     }
}
```

this way we only use GetName() and not assume a correlation between Name and GetName() (same for description)

wazzamatazz: Okay, will do.

wazzamatazz: Done: https://github.com/RicoSuter/NJsonSchema/pull/1313/commits/3aa50c8ced3fb51c89f648991bbb623d7b94b907

RicoSuter:
Thx lgtm

PR #1317, opened 2021-02-17
Fix in JsonReferenceResolver for referencing definitions in other files

twodogmbh:


src/NJsonSchema/JsonReferenceResolver.cs
```diff
@@ -189,7 +189,7 @@ private async Task<IJsonReference> ResolveFileReferenceWithAlreadyResolvedCheckA
                 if (resolvedSchema is JsonSchema && append &&
                     (_schemaAppender.RootObject as JsonSchema)?.Definitions.Values.Contains(referencedFile) != true)
                 {
-                    var key = arr[0].Split('/', '\\').Last().Split('.').First();
+                    var key = jsonPath.Split('/', '\\').Last().Split('.').First();
```
RicoSuter, line 192: Are you sure it was a bug to use `arr[0]` and not `jsonPath`? 
I'm a bit afraid that this will fix your problem but introduce many other regressions.

twodogmbh: It was `jsonPath` before pull request #1258 and we were able to generate C# code out of quite complex JSON schemas.
Now we wanted to update to a newer version where we no longer can generate code because of this issue.

When you define a schema with a property reference to another file like "$ref" : "./Example.json#/definitions/Foo". The key in the JsonSchema.Definititions dictionary is set to "Example" instead of "Foo", which leads to a property of type "Example" instead of "Foo", because 'arr[0]' contains only the path up to the filename.

RicoSuter: Aha ok, makes sense to me as this seems to be a regression introduced by #1258...
/cc @shadow-cs any comments on this before we merge?

shadow-cs: I think @twodogmbh is right and we introduced a bug there when we blindly replaced `jsonPath` with `arr` even at places where we shouldn't.

RicoSuter: Ok then, do we need other changes or is the PR ready to merge?

twodogmbh: I would appreciate if you could merge this PR so we can update to a newer version of NJsonSchema, as this bug is currently blocking us. Thanks

RicoSuter:
https://github.com/RicoSuter/NJsonSchema/pull/1318

PR #1319, opened 2021-02-22
Issue in generation of C# classes with inheritance fixed

twodogmbh:
Issue in the `JsonSchemaAppender` for the generation of classes fixed when the definitions reference the root schema. When the passed schema is already the root schema the method returns immediately instead of throwing an exception. In this case the definitions are already known and there is no reason to throw an exception.

Issue in the `CSharpTypeResolver` for the generation of base classes without properties fixed. When classes don't contain properties but contain the "discriminator" attribute they should still be generated. Otherwise it's not possible to define a class hierarchy where the parent class does not contain any property.
My assumption is that when "discriminator" is defined there is at least one derived class which is interested in this specific type, so it would be wrong to just replace it with AnyType (object)

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp.Tests/InheritanceTests.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp.Tests/InheritanceTests.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
+using System.IO;
+using System.Reflection;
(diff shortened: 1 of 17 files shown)


twodogmbh:
@RicoSuter Please can you review and merge the changes, if you agree with them. Thank you

RicoSuter:
Lgtm, thanks for the PR. 
Your description makes sense but I'm still a bit afraid :-)

PR #1320, opened 2021-02-22
Add schema title property to ClassTemplateModelBase

mmdtgdev:
I added a property "SchemaTitle" to ClassTemplateModelBase. The reason is that the "ClassName" is not the original one from the swagger file any more (at least the first character is always capitalized). We need the original unchanged title from the Swagger file in the class.liquid file so that we can add a custom attribute with it to the resulting code file on generation.

### DIFF
--- a/src/NJsonSchema.CodeGeneration/Models/ClassTemplateModelBase.cs
+++ b/src/NJsonSchema.CodeGeneration/Models/ClassTemplateModelBase.cs
@@ -37,7 +37,7 @@ namespace NJsonSchema.CodeGeneration.Models
         /// <summary>
         /// Gets the original title of the class (schema title).
         /// </summary>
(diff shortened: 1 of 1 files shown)


RicoSuter:
Does it have to be virtual?

mmdtgdev:
You're right - it's not absolutely necessary. I committed another change making it non-virtual. (I just made it virtual to provide the ability to override it in derived classes if needed)

mmdtgdev:
@RicoSuter I don't know why the tests have failed after this very simple change (making prop virtual -> non-virtual) - maybe an one-time error in Azure DevOps? Maybe it's possible for you to manually reschedule a check-run? Thank you very much in advance!

RicoSuter:
Was a problem with benchmarks (slow DevOps). All fine. Thanks for the PR.

PR #1321, opened 2021-02-22
Fixed typo in template for Text.Json.Serialization.JsonPropertyName

Voronkov-A:
Just a small fix of typo, introduced by [this](https://github.com/RicoSuter/NJsonSchema/commit/19a01a0d38e1a919f658f50229a0a28c6c316897) commit, that leads to code like `[System.Text.Json.Serialization.JsonPropertyName("type"]` and compilation errors.


RicoSuter:
Btw, STJ support is just at the beginning (experimental) and some things might not work yet as expected (eg in NSwag I did not even start to using it).

Voronkov-A:
> Btw, STJ support is just at the beginning (experimental) and some things might not work yet as expected (eg in NSwag I did not even start to using it).

I see, thanks. Though, it is enough for what I need now :)

Issue #1322, opened 2021-02-23
Question: Reasons to choose format of "guid" over "uuid" for string?


martin-shields-sage:
I have a question, I see that the JsonFormatStrings has an obsolete value for UUID and states to use GUID instead as uuid is redundant. However, if I have a JSON schema of..

```
"SubType" : {
              "type": "object",
              "properties": {
                "Id": {
                  "type": "string",
                  "format": "uuid"
                }
              },
              "required": [
                "Id"
              ]
            }
...
```

This does not get validated as a UUID or GUID type when running the validation against this schema without putting a new custom validator in to add UUID. Was there any reason why this was made redundant? When I look at the JSON schema specification I still see links to UUID being there - https://json-schema.org/draft/2020-12/json-schema-validation.html

RicoSuter:
I think even if UUID is obsolete we should support it in the validator: 
https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema/Validation/JsonSchemaValidator.cs#L28

martin-shields-sage:
> I think even if UUID is obsolete we should support it in the validator:
> https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema/Validation/JsonSchemaValidator.cs#L28

I would agree, it looks like the line which you have highlighted does not validate the example I have provided above (being a uuid instead of guide)?

martin-shields-sage:
Would you suggest that we add a new format validator to the list based on https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema/Validation/JsonSchemaValidator.cs#L28?

RicoSuter:
> Would you suggest that we add a new format validator to the list based on

Yes. Can you create a PR?

martin-shields-sage:
> > Would you suggest that we add a new format validator to the list based on
> 
> Yes. Can you create a PR?

Sure.

PR #1325, opened 2021-02-24
Resolves  RicoSuter/NJsonSchema#1322 - Added uuid format for schema validation

martin-shields-sage:
- Created new uuid format validator
- Added testing for format

Resolves RicoSuter/NJsonSchema#1322


RicoSuter:
Thanks for the PR.

PR #1331, opened 2021-02-27
Fix supporting System.Text.Json in CSharpJsonSerializerGenerator

unchase:
Added the tests to **CSharpJsonSerializerGeneratorTests**.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/CSharpGeneratorSettings.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp/CSharpGeneratorSettings.cs
@@ -121,11 +121,8 @@ namespace NJsonSchema.CodeGeneration.CSharp
         /// <summary>Gets or sets a value indicating whether to use preserve references handling (All) in the JSON serializer (default: false).</summary>
         public bool HandleReferences { get; set; }
 
-        /// <summary>Gets or sets the name of a static method which is called to transform the JsonSerializerSettings (Newtonsoft.Json) used in the generated ToJson()/FromJson() methods (default: null).</summary>
-        public string JsonSerializerSettingsTransformationMethod { get; set; }
-
-        /// <summary>Gets or sets the name of a static method which is called to transform the JsonSerializerOptions (System.Text.Json) used in the generated ToJson()/FromJson() methods (default: null).</summary>
-        public string JsonSerializerOptionsTransformationMethod { get; set; }
+        /// <summary>Gets or sets the name of a static method which is called to transform the JsonSerializerSettings (for Newtonsoft.Json) or the JsonSerializerOptions (for System.Text.Json) used in the generated ToJson()/FromJson() methods (default: null).</summary>
+        public string JsonSerializerSettingsOrOptionsTransformationMethod { get; set; }
 
         /// <summary>Gets or sets a value indicating whether to render ToJson() and FromJson() methods (default: true).</summary>
         public bool GenerateJsonMethods { get; set; }
--- a/src/NJsonSchema.CodeGeneration.CSharp/CSharpGeneratorSettings.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp/CSharpGeneratorSettings.cs
@@ -57,7 +57,7 @@ namespace NJsonSchema.CodeGeneration.CSharp
         /// <summary>Gets or sets the .NET namespace of the generated types (default: MyNamespace).</summary>
         public string Namespace { get; set; }
 
-        /// <summary>Gets or sets a value indicating whether a required property must be defined in JSON 
+        /// <summary>Gets or sets a value indicating whether a required property must be defined in JSON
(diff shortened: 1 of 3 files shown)


unchase:
@RicoSuter ready for review.

src/NJsonSchema.CodeGeneration.CSharp/CSharpGeneratorSettings.cs
```diff
@@ -121,9 +121,12 @@ public CSharpGeneratorSettings()
         /// <summary>Gets or sets a value indicating whether to use preserve references handling (All) in the JSON serializer (default: false).</summary>
         public bool HandleReferences { get; set; }
 
-        /// <summary>Gets or sets the name of a static method which is called to transform the JsonSerializerSettings used in the generated ToJson()/FromJson() methods (default: null).</summary>
+        /// <summary>Gets or sets the name of a static method which is called to transform the JsonSerializerSettings (Newtonsoft.Json) used in the generated ToJson()/FromJson() methods (default: null).</summary>
         public string JsonSerializerSettingsTransformationMethod { get; set; }
```
RicoSuter, line 125: Maybe we should only use this method and use it for both cases just to avoid having mutual-exclusive options and avoiding adding more options... downside is the naming but what do you think?

RicoSuter:
Thanks for adding also some unit tests.. was too lazy, but it's important to avoid regressions.

src/NJsonSchema.CodeGeneration.CSharp/CSharpJsonSerializerGenerator.cs
```diff
@@ -23,47 +23,90 @@ public static string GenerateJsonSerializerParameterCode(CSharpGeneratorSettings
...
+                case CSharpJsonLibrary.SystemTextJson:
+                    var useOptionsTransformationMethod = !string.IsNullOrEmpty(settings.JsonSerializerOptionsTransformationMethod);
+                    // TODO: add more conditions?
+                    if (useOptionsTransformationMethod || hasJsonConverters)
+                    {
+                        return ", " +
+                            (useOptionsTransformationMethod ? settings.JsonSerializerOptionsTransformationMethod + "(" : string.Empty) +
+                            "System.Text.Json.JsonSerializerOptions { " +
+
+                            // TODO(system.text.json): Add supporting more options for System.Text.Json (hided for Newtonsoft.Json)
```
RicoSuter, line 74: I think we should only support the existing "high-level" configs which already exists - if you need more you can use JsonSerializerOptionsTransformationMethod

src/NJsonSchema.CodeGeneration.CSharp/CSharpGeneratorSettings.cs
unchase: I think we can use it property for both json libraries. Just will rename it to `JsonSerializerSettingsOrOptionsTransformationMethod`. Do you agree?

src/NJsonSchema.CodeGeneration.CSharp/CSharpJsonSerializerGenerator.cs
unchase: Ok, I'll just leave these options.

src/NJsonSchema.CodeGeneration.CSharp/CSharpGeneratorSettings.cs
```diff
@@ -121,8 +121,8 @@ public CSharpGeneratorSettings()
         /// <summary>Gets or sets a value indicating whether to use preserve references handling (All) in the JSON serializer (default: false).</summary>
         public bool HandleReferences { get; set; }
 
-        /// <summary>Gets or sets the name of a static method which is called to transform the JsonSerializerSettings used in the generated ToJson()/FromJson() methods (default: null).</summary>
-        public string JsonSerializerSettingsTransformationMethod { get; set; }
+        /// <summary>Gets or sets the name of a static method which is called to transform the JsonSerializerSettings (for Newtonsoft.Json) or the JsonSerializerOptions (for System.Text.Json) used in the generated ToJson()/FromJson() methods (default: null).</summary>
+        public string JsonSerializerSettingsOrOptionsTransformationMethod { get; set; }
```
RicoSuter, line 125: I think we should just use JsonSerializerSettingsTransformationMethod to avoid breaking changes and having only one config which applies for any target json library... even if the name is not 100% "clean".

unchase: I am agree with you. Fixed)

RicoSuter:
Thx for the PR.

unchase:
You are welcome!
@RicoSuter could you please to publish changes to the new release (v10.3.10) that I will use it for try to solve [NSwag#2243](https://github.com/RicoSuter/NSwag/issues/2243)

Issue #1328, opened 2021-02-26
JsonSchemaGenerator generate schema unexpectedly!!!Why not obay the DataContract?


mokeyish:
https://github.com/RicoSuter/NJsonSchema/blob/d8f5302ed5315167c361358735a90718de7abcd0/src/NJsonSchema/Generation/JsonSchemaGenerator.cs#L901

JsonSchemaGenerator generate schema unexpectedly!!!Why not obay the DataContract?

Here is my case code:
```c#

[DataContract]
class A {
     [IgnoreDataMember]
     public string MyField{get => _myField;
     set {
            // Do some stuff, excep for deserializing.
            _myField = value;
           }
    } 

     [DataMember(Name="MyField")]
     [Display(Name="My Field", Description="......")]
     private string _myField;
}
```

**NJsonSchema** cannot gernerate property `MyField` for` class A`;

RicoSuter:
Can you create a PR with a test covering this misbehavior and also covering regular serialization/deserialization?

PR #1335, opened 2021-03-04
When class with [DataContract], field or property with [DataMember] is not ignored.

killyu:
https://github.com/RicoSuter/NJsonSchema/issues/1328

### DIFF
--- a/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
+++ b/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
@@ -20,6 +20,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.Serialization;
 
 namespace NJsonSchema.Generation
 {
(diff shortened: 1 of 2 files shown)


src/NJsonSchema/Generation/JsonSchemaGenerator.cs
```diff
@@ -858,12 +859,13 @@ private void GenerateProperties(Type type, JsonSchema schema, JsonSchemaResolver
...
                 .DeclaredFields
-                .Where(f => !f.IsPrivate && !f.IsStatic)
+                .Where(f => !f.IsPrivate && !f.IsStatic || f.IsDefined(typeof(DataMemberAttribute)))
                 .OfType<MemberInfo>()
                 .Concat(
                     type.GetTypeInfo().DeclaredProperties
                     .Where(p => (p.GetMethod?.IsPrivate != true && p.GetMethod?.IsStatic == false) ||
-                                (p.SetMethod?.IsPrivate != true && p.SetMethod?.IsStatic == false))
+                                (p.SetMethod?.IsPrivate != true && p.SetMethod?.IsStatic == false) ||
+                                p.IsDefined(typeof(DataMemberAttribute)))
```
RicoSuter, line 868: Can you also implement this for the "legacy" part?

killyu: Nonlegacy is OK. But after I change the NJsonSchema project target to net40, I can't debug the project for the "legacy" part. Even if I create a new net40 console project for test, there is a weird [CS0103] error.

RicoSuter: Ok then it's fine.

PR #1336, opened 2021-03-04
root element of external schema ref files cannot be same name as any property of JsonSchema

jimfrederic:
No solution, just a failing test case to reproduce the problem.

### DIFF
--- a/src/NJsonSchema.Yaml.Tests/NJsonSchema.Yaml.Tests.csproj
+++ b/src/NJsonSchema.Yaml.Tests/NJsonSchema.Yaml.Tests.csproj
@@ -5,7 +5,7 @@
   </PropertyGroup>
   <PropertyGroup>
       <DocumentationFile>bin\Debug\$(TargetFramework)\NJsonSchema.Yaml.Tests.xml</DocumentationFile>
-  </PropertyGroup>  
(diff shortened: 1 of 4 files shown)


RicoSuter:
`schema_named_title.yaml` is not a valid JSON Schema... 
is this even supposed to work?

jimfrederic:
> `schema_named_title.yaml` is not a valid JSON Schema...
> Is this even supposed to work (ie. should this be allowed)?

The example is contrived, admittedly, cut down from a real-world OpenApi document to minimize it.  So take a similar example: If you modify an existing test, changing `myBool` to `Title` [here in collection.json](https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema.Tests/References/LocalReferencesTests/collection.json#L5) and [here in schema_with_collection_reference.json](https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema.Tests/References/LocalReferencesTests/schema_with_collection_reference.json#L5), then the test case [When_schema_references_collection_in_definitions_it_works](https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema.Tests/References/LocalReferencesTests.cs#L43) throws the same exception while parsing.

PR #1339, opened 2021-03-08
Fix ToJson and FromJson liquid files for System.Text.Json

unchase:
image

Now generates:

```csharp
...
public string ToJson() 
{
    return System.Text.Json.JsonSerializer.Serialize(this, new System.Text.Json.JsonSerializerOptions(); var converters = new System.Text.Json.Serialization.JsonConverter[] { new Converter1() });
}
...
```

image

---

Fixed in this PR (need to check).

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/Templates/Class.FromJson.liquid
+++ b/src/NJsonSchema.CodeGeneration.CSharp/Templates/Class.FromJson.liquid
@@ -1,7 +1,7 @@
 ﻿public static {{ ClassName }} FromJson(string data)
 {
 {% if UseSystemTextJson -%}
-    return System.Text.Json.JsonSerializer.Deserialize<{{ ClassName }}>(data{{ JsonSerializerParameterCode }});
+    return System.Text.Json.JsonSerializer.Deserialize<{{ ClassName }}>(data{{ JsonSerializerParameterCode | split: ';' | first }});
 {% else -%}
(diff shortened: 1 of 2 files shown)


unchase:
@RicoSuter ready for review

unchase:
@RicoSuter ready for review

RicoSuter:
In the end we should clean this up as I'm not a fan to parse/split etc. code in the templates.. but for now it's probably fine. What do you think=

jeremyVignelles:
I might have missed something, but what's the rationale of this? why are there two instructions that you need to split?
Why should it be done there rather than in a property in the C# code?

jeremyVignelles:
It seems that the `CSharpJsonSerializerGenerator.GenerateJsonSerializerParameterCode` is used for the purpose of setting the parameters to the Deserialize and Serialize methods. Why is there a `; var converters =` there ?

> Refactor the CSharpJsonSerializerGenerator.GenerateForJsonLibrary method in such a way that backward compatibility is lost (due to the possible addition of a separate method that returns the addition of custom converters for System.Text.Json).

I'd be OK for breaking backward compatibility to add a new Method/Property that would be specific to STJ

That said, I don't have the whole picture and that's still quite obscure for me, but making assumption on the shape of the generated code seems really rigid to me. What if I want to override the C# code to generate the converters my own way ?

unchase:
Need to think about how to refactor the method to make it more flexible.

unchase:
Related: #1341

PR #1346, opened 2021-03-25
Replace DotLiquid with Fluid

lahma:


microalps:
Just out of curiousity, we made some changes to the parser in dotliquid. How does that benchmark for you? There are more improvements to be made but an idea of what needs to be done to keep DotLiquid would be good. Try compiling latest with this patch https://github.com/dotliquid/dotliquid/pulls/422

lahma:
@microalps honestly I think it would serve DotLiquid best to have a benchmark suite in its own repository. I would think @sebros wouldn't mind if you use Fluid's comparison suite as starting point, you get [a nice comparison report](https://github.com/sebastienros/fluid#performance) with it.

sebastienros:
@microalps have you looked at the benchmarks on the Fluid's page? It has DotLiquid ones. You can clone them and reference your fork to run them. But I doubt you would be able to "change" DL to the point it's comparable. At least for rendering it should be possible since it's less intrusive, no Regex should be necessary.

RicoSuter:
Thanks for this PR - looks promising.
It's just important that we do not break anything (keep same output). 
Will you also update the NSwag template as soon as this NJS version is updated there?
Apart from that do we need to do any code changes in NSwag?

lahma:
> It's just important that we do not break anything (keep same output).

Currently the "regressions" I see are some whitespace changes on NSwag side, like newlines and like removed whitespace for empty lines. These can be tweaked on NSwag side. I have already a branch that has template changes that fix most of the findings I have against TS/C# generation.

> Will you also update the NSwag template as soon as this NJS version is updated there?

I'm committed to create a PR to integrate with NSwag and tweak until acceptable. Templates require mostly changes like from `-%}` to `{%-` usage. Some newlines etc.

> Apart from that do we need to do any code changes in NSwag?

Biggest change is the target frameworks, issue here: https://github.com/RicoSuter/NSwag/issues/3384 . Easier maybe to do in iterations if we have open communication line to get things rolling.

Here's the current diff for changes: https://github.com/RicoSuter/NSwag/compare/master...lahma:fluid?expand=1

lahma:
I've rebased against master and updated the benchmarks (master with new Namotion.Reflection and this branch with new Namotion.Reflection).

src/NJsonSchema/NJsonSchema.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;netstandard2.0;net40;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
```
RicoSuter, line 3: This is a bit a problem as i know there are ppl using this lib with netstd < 2.0 and also with .NET 4.0
Is this really required?

lahma: I restored the old targets, seems that we can keep this project targeting as-is. Good catch.

lahma:
Converted to draft as I still need to one bigger round verifying the generated output (both our internal code base and NSwag outputs).

RicoSuter:
> didn't have correct whitespace handling for stripping trailing whitespace

So Fluid correctly handles this as specified by Shopify?

> Converted to draft as I still need to one bigger round verifying the generated output

Please ping me when it's ready then ill have a look and we can continue integrating this (eg in NSwag).

lahma:
> > didn't have correct whitespace handling for stripping trailing whitespace
> 
> So Fluid correctly handles this as specified by Shopify?

Yes, Fluid is very much stricter and follows the Shopify spec by the dot, correct me if I'm wrong @sebastienros .

> > Converted to draft as I still need to one bigger round verifying the generated output
> 
> Please ping me when it's ready then ill have a look and we can continue integrating this (eg in NSwag).

I'll work with these, the latest changes in master just broke things a bit more, I'll probably submit a PR to fix regression and new performance issue.

sebastienros:
It follows Shopify as much as it can. And uses options to diverge. Same for filters, all based on specs, tests or manual testing on shopify.com directly.

lahma:
Seems that upgrading to latest Namotion.Reflection fixed the test problems, back to the original quest...

RicoSuter:
Any updates on this PR?

lahma:
Running with latest release against our source code tree (TS and C# client generation)

```
NSwag command line tool for .NET Core Net50, toolchain v13.13.2.0 (NJsonSchema v10.5.2.0 (Newtonsoft.Json v12.0.0.0))
Visit http://NSwag.org for more information.
NSwag bin directory: C:\Users\marko.lahma\.nuget\packages\nswag.msbuild\13.13.2\tools\Net50

Executing file 'nswag.json' with variables 'Configuration=Debug'...
Done.

Duration: 00:00:31.8092628
```

With migrated code base using Fluid:

```
NSwag command line tool for .NET Core Net50
Visit http://NSwag.org for more information.
NSwag bin directory: C:\Work\OpenSource\NSwag\src\NSwag.ConsoleCore\bin\Release\net5.0

Executing file 'nswag.json' with variables 'Configuration=Debug'...
Done.

Duration: 00:00:07.9538140
```

Probably half or even more of the 8 seconds is getting the OpenAPI document.

lahma:
@RicoSuter I have locally updated NSwag templates and generated code is almost the same, slight differences with line breaks and trimming, but nothing too serious I hope. This is a big change so regressions probably are inevitable, but we'll see... Ready for review.

PR #1347, opened 2021-03-26
Create benchmarks

lahma:
As I've worked with Fluid integration I've also created a baseline of benchmarks to tests against to validate improvements. I think it would be beneficial to have this in repo to aid with other PRs (I've found some other performance problems besides the template library).

Using BenchmarkDotNet with memory diagnoser to get trustworthy reports easily via command line, `dotnet run -c Relese` for benchmark projects. The xUnit test cases still seem to work as expected at least with Rider.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ src/packages/Newtonsoft.Json**
 **.sln.ide/**
 **.vs/**
 .vscode/
+.idea/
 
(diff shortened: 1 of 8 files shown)


RicoSuter:
Is this ready to be merged?

lahma:
Yes please! We can add more benchmarks later for new areas if needed.

RicoSuter:
Thx for the PR.

PR #1349, opened 2021-03-26
Reduce memory allocations

lahma:
* avoid enum HasFlag as it allocates
* avoid interface dispatch for collections that are accessed a lot
* remove some LINQ that slows things down
* avoid delegate allocations for each collection setup, use delegate stored in a field (preserves `this` context)
* avoid string.Split(params char[]) trap by supplying fixed char array

### DIFF
--- a/src/NJsonSchema/Visitors/AsyncJsonReferenceVisitorBase.cs
+++ b/src/NJsonSchema/Visitors/AsyncJsonReferenceVisitorBase.cs
@@ -124,22 +124,22 @@ namespace NJsonSchema.Visitors
                     await VisitAsync(schema.Items.ElementAt(i), path + "/items[" + i + "]", null, checkedObjects, o => ReplaceOrDelete(schema.Items, index, (JsonSchema)o), cancellationToken).ConfigureAwait(false);
                 }
 
-                for (var i = 0; i < schema.AllOf.Count; i++)
+                for (var i = 0; i < schema._allOf.Count; i++)
                 {
                     var index = i;
-                    await VisitAsync(schema.AllOf.ElementAt(i), path + "/allOf[" + i + "]", null, checkedObjects, o => ReplaceOrDelete(schema.AllOf, index, (JsonSchema)o), cancellationToken).ConfigureAwait(false);
+                    await VisitAsync(schema._allOf.ElementAt(i), path + "/allOf[" + i + "]", null, checkedObjects, o => ReplaceOrDelete(schema.AllOf, index, (JsonSchema)o), cancellationToken).ConfigureAwait(false);
                 }
 
-                for (var i = 0; i < schema.AnyOf.Count; i++)
+                for (var i = 0; i < schema._anyOf.Count; i++)
                 {
                     var index = i;
-                    await VisitAsync(schema.AnyOf.ElementAt(i), path + "/anyOf[" + i + "]", null, checkedObjects, o => ReplaceOrDelete(schema.AnyOf, index, (JsonSchema)o), cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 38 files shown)


src/NJsonSchema/Visitors/AsyncJsonReferenceVisitorBase.cs
```diff
@@ -136,10 +136,10 @@ protected virtual async Task VisitAsync(object obj, string path, string typeName
                     await VisitAsync(schema.AnyOf.ElementAt(i), path + "/anyOf[" + i + "]", null, checkedObjects, o => ReplaceOrDelete(schema.AnyOf, index, (JsonSchema)o), cancellationToken).ConfigureAwait(false);
                 }
 
-                for (var i = 0; i < schema.OneOf.Count; i++)
+                for (var i = 0; i < schema._oneOf.Count; i++)
```
RicoSuter, line 139: Why do you change that? 
Is there a perf difference? It should be inlined in the old form, no?

lahma: (Here's IL comparison)[https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAmARgFgAoQgBgAJDSUBuK2h0gOgGEJdcYYDNggA7AM6cA8sABWgjAFkIAExi5WlNgGYGxOtzoBvKnTN1T5gMpgAFjAC2AQzri7jlwF46ogK79NcwtKIMJdQhQ6AElRDBgoADMnMBgACgBKYKCTEKDzBOg6VIA3Jyg6bDpvGmYKugAeV3dnaVEYKQSeCF9Y2uwAan70yzzjEbyAX3GprVyzMIZI3lEwWDiMrPMc0bMC8pKyuuq+hqb7FoB9MXbO3h6MPsHhubztnaDxoJmv8d+XhYiBjEqxgcQud1i3GS9hUG3Gb1GpXKkHuVTOHk4VzaHS690COz2RSRRzoNTqjRRvQqT0+W1po3pdG+5hmVFZ1HC+hs5ycVARFVi8VETlwdBk4nipWAAl4/AUIlE9UYNAAfHQsTd8WZxgsorKBEIFUrSKqxdiEvDGQBzUFVNUajpapkfSgTIA] so as seemed to be in hot path (with Fluid as actually can see something in profiler outside of DotLiquid). The samples show first how two virtual calls in loop, then one with new code and third alternative where no virtual calls in loop. 

This is still work in progress and hard to see difference against current master as all time is spent in DotLiquid.

lahma: And by all means this is not the biggest thing to consider with. I just tend to see it happen a lot of libraries that fields can be declared as `IDictionary<string, object> _theField = new Dictionary<string, object>();` and after that everything is virtual dispatch slowing things down. [I worked extensively with Jint](https://github.com/sebastienros/jint/issues/451#issuecomment-412627408) optimizing all kinds of access patterns.

RicoSuter: > This is still work in progress and hard to see difference against current master as all time is spent in DotLiquid.

So we'd first move to Fluid without this PR and then do optimizations on top of Fluid?

lahma: Ideally if possible, Fluid makes it *really fast* compared to baseline and that would be come the new baseline which would be easier to do targeted optimizations against.

RicoSuter:
Any updates on this PR?

lahma:
I still need to find time to validate this after rebases.

lahma:
OK I'm pretty happy with this now. Just by reducing memory pressure and some faster paths using concrete types instead of interface dispatch we get a nice performance boost.

Running with custom NSwag (referencing local NJsonSchema) and Namoting.Reflection 2.0.6. Console core against our product running with NET 6.0 runner. Fastest result from 10 runs on my laptop. No regressions in client generation.

Master: `Duration: 00:00:07.4000759`

This PR: `Duration: 00:00:06.4253551`

PR #1353, opened 2021-04-07
Make the GetKey protected & virtual to allow overriding it

mikoskinen:
Hi,

In this change the JsonSchemaResolver's GetKey is made protected & virtual in order to use additional information than just the FullName as the mapping key. In our case we would like to use Assembly version in addition to fullname.

In our scenario controllers are loaded runtime and there might be two classes with the exact same FullName coming from two different assemblies and we need to generate schemas for both of them as the types may be different (different properties etc.).

### DIFF
--- a/src/NJsonSchema/Generation/JsonSchemaResolver.cs
+++ b/src/NJsonSchema/Generation/JsonSchemaResolver.cs
@@ -68,7 +68,13 @@ namespace NJsonSchema.Generation
         /// <summary>Gets all the schemas.</summary>
         public IEnumerable<JsonSchema> Schemas => _mappings.Values;
 
-        private string GetKey(Type type, bool isIntegerEnum)
+        /// <summary>
+        /// Gets the mapping key for the given type.
+        /// </summary>
+        /// <param name="type">The type.</param>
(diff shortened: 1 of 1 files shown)


RicoSuter:
Shouldnt we fix that directly instead of externally? ie. change the current impl to fix your issue?

mikoskinen:
Thanks for the comment! The idea was to break as little as possible :) This way the developer can override the JsonSchemaResolver (or OpenApiSchemaResolver) and provide their own implementation. 

Using the Type.Assembly and its version in addition to Type.FullName isn't always possible as the assemblies can be dynamic. For example in one of our cases we are generating the assemblies on the fly using Roslyn and the assembly is memory only and missing the version details. In this case the current implementation works fine as we must rely on the Type.FullName.

In another case it may be that we use the Type's FullName to map it to some other non-type related data, for example to Nuget package. In this scenario the application may have loaded two versions of the same Nuget-package as plugins (using Plugin Framework) and even though both packages have different versions, the assemblies inside them may have the same 1.0.0 version. We need some application specific mechanism for deciding if we expect the schema's to be the same or not.

RicoSuter:
Renamed to GetTypeKey

PR #1357, opened 2021-04-09
cast dictionary to any during assignment

jjoekoullas:
This attempts to fix https://github.com/RicoSuter/NSwag/issues/2881 by simply casting dictionaries to any before doing assignment.

As far as I can tell, there isn't a great way to fix this outside of casting the dictionary to any.  Even if the keys are properly typed, you can still run into issues with read-only values being assigned.  Since this is a fully internal operation casting the dictionary to any seems safe enough.

I'm having a lot of trouble figuring out how to get this built and how to use it locally in order to test it however.


RicoSuter:
As you only have changes in templates, you can test your changes by simply overriding the given templates: 
https://github.com/RicoSuter/NJsonSchema/wiki/Templates

If everything is fine we can merge the PR.

jjoekoullas:
@RicoSuter thank you!

Tested my changed via the template directory like you suggested - they're working fine now.

Updated tests, they pass locally.

RicoSuter:
Thanks for the PR.

PR #1358, opened 2021-04-18
Add typeName to ArgumentException message in CodeArtifact

thomasjoscht:
The name of type which leads to ArgumentException in CodeArtifact ctor can help to analyse problems in invalid schema. 

### Background
Observed this ArgumentException in Visual Studio. (CodeGeneration with Unchase plugin). Analyzing reason was very difficult because of missing type name which was invalid in my big schema. This patch helps to find the concerned part in schema.


RicoSuter:
Thx for the PR.

PR #1360, opened 2021-04-24
Arrays and dictionaries in class DTOs now default to the correct null value in the init() method

### LINKED ISSUE #1359: Typescript code generation incorrect for classes with nullable arrays and NullValue = TypeScriptNullValue.Null
Here's an example of class code generated for a model with a nullable array property:

Generator settings:
```cs
NullValue = TypeScriptNullValue.Null,
TypeScriptVersion = 4,
SchemaType = SchemaType.OpenApi3,
MarkOptionalProperties = false,
```

Schema:
```cs
var schema = new JsonSchema
{
    Properties =
    {
        { "A", new JsonSchemaProperty
            {
                Type = JsonObjectType.String,
                IsRequired = true,
                IsNullableRaw = true
            }
        },
        { "B", new JsonSchemaProperty
            {
                Type = JsonObjectType.Array,
                Item = new JsonSchema
                {
                    Type = JsonObjectType.String
                },
                IsRequired = true,
                IsNullableRaw = true
            }
        },
    }
};
```

Generated code:
```ts
export class MyClass implements IMyClass {
    a!: string | null;
    b!: string[] | null;

    constructor(data?: IMyClass) {
        if (data) {
            for (var property in data) {
                if (data.hasOwnProperty(property))
                    (<any>this)[property] = (<any>data)[property];
            }
        }
    }

    init(_data?: any) {
        if (_data) {
            this.a = _data["A"] !== undefined ? _data["A"] : <any>null;
            if (Array.isArray(_data["B"])) {
                this.b = [] as any;
                for (let item of _data["B"])
                    this.b!.push(item);
            }
        }
    }

    static fromJS(data: any): MyClass {
        data = typeof data === 'object' ? data : {};
        let result = new MyClass();
        result.init(data);
        return result;
    }

    toJSON(data?: any) {
        data = typeof data === 'object' ? data : {};
        data["A"] = this.a !== undefined ? this.a : <any>null;
        if (Array.isArray(this.b)) {
            data["B"] = [];
            for (let item of this.b)
                data["B"].push(item);
        }
        return data; 
    }
}

export interface IMyClass {
    a: string | null;
    b: string[] | null;
}
```

The types generated are correct, however the implementation of `init()` is slightly incorrect.
When `fromJS` is called with this object for example:

```js
{
    a: null;
    b: null;
}
```

The resulting class will have properties
```js
{
    a: null;
    b: undefined;
}
```

Note that property `b` is `undefined` instead of `null` as expected.

JustASquid:
Fixes #1359 

This is my first PR here, so feedback is very welcome.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.TypeScript/Templates/ConvertToClass.liquid
+++ b/src/NJsonSchema.CodeGeneration.TypeScript/Templates/ConvertToClass.liquid
@@ -21,7 +21,7 @@ if (Array.isArray({{ Value }})) {
 {% if NullValue != "undefined" %}else {
     {{ Variable }} = <any>{{ NullValue }};
 }
-{% endif %}
+{% endif -%}
 {% elseif IsDictionary -%}
 if ({{ Value }}) {
     {{ Variable }} = {} as any;
@@ -45,7 +45,7 @@ if ({{ Value }}) {
 {% if NullValue != "undefined" %}else {
     {{ Variable }} = <any>{{ NullValue }};
 }
-{% endif %}
+{% endif -%}
 {% else -%}
     {% if IsDate or IsDateTime -%}
 {{ Variable }} = {{ Value }} ? {{ StringToDateCode }}({{ Value }}.toString()) : {% if HasDefaultValue %}{{ StringToDateCode }}({{ DefaultValue }}){% else %}<any>{{ NullValue }}{% endif %};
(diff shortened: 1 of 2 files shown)


RicoSuter:
In general this looks fine (didnt touch this for months). Maybe you can also create a unit test verifying the output?

src/NJsonSchema.CodeGeneration.TypeScript/Templates/ConvertToClass.liquid
```diff
@@ -37,6 +39,8 @@ if ({{ Value }}) {
 {%         endif -%}
 {%     endif -%}
     }
+} else {
```
RicoSuter, line 42: Wouldnt it be better to use elseif nullvalue != undefined to avoid generating some unused code?

JustASquid: Good point - updated.

PR #1367, opened 2021-05-08
System.Text.Json.Serialization.JsonExtensionDataAttribute allows object & JsonElement values

deceptiveSimplicity:
According to https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to?pivots=dotnet-5-0#deserialization-of-object-properties System.Text.Json deserializes to JsonElement even if the property is of type IDictionary<string, object>.

### DIFF
--- a/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
+++ b/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
@@ -767,8 +767,6 @@ namespace NJsonSchema.Generation
             {
                 var genericTypeArguments = extensionDataProperty.GenericArguments;
                 var extensionDataPropertyType = genericTypeArguments.Length == 2 ? genericTypeArguments[1] : typeof(object).ToContextualType();
-                if (extensionDataPropertyType.TypeInfo.FullName.Equals("System.Text.Json.JsonElement"))
-                    extensionDataPropertyType = typeof(object).ToContextualType();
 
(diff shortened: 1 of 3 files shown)


src/NJsonSchema/Generation/JsonSchemaGenerator.cs
```diff
@@ -767,6 +767,8 @@ private void ApplyAdditionalProperties<TSchemaType>(TSchemaType schema, Type typ
             {
                 var genericTypeArguments = extensionDataProperty.GenericArguments;
                 var extensionDataPropertyType = genericTypeArguments.Length == 2 ? genericTypeArguments[1] : typeof(object).ToContextualType();
+                if (extensionDataPropertyType.TypeInfo.FullName.Equals("System.Text.Json.JsonElement"))
+                    extensionDataPropertyType = typeof(object).ToContextualType();
```
RicoSuter, line 771: I think its better if we handle this here:
https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema/Generation/DefaultReflectionService.cs#L177

This would ensure that its always handled as object not only in the case of extension data.. also is element an object ({}) or any (also primitive value allowed).

deceptiveSimplicity: Moved the check to DefaultReflectionService.

PR #1372, opened 2021-05-24
format new lines in C# types with a <br/> tag

adamjones1:


RicoSuter:
Thanks for the PR. 
Makes sense to add... 
the optimal solution would be to use a full fledged markdown to HTML converter :)

PR #1377, opened 2021-06-16
Fix XML comments on CodeGeneratorSettings

scharnyw:
Fix some inconsistencies in the documented default values in XML comments of `CSharpGeneratorSettings` and `TypeScriptGeneratorSettings`.


RicoSuter:
Thanks for the PR.

PR #1382, opened 2021-06-25
Improve LeafType generated code

trejjam:
The generated code using the `UseLeafType` was not 100% correct for arrays. This PR changes generated code from:
```ts
property: A[] | B[],
```
to:
```ts
property: (A | B)[],
```


RicoSuter:
Thanks for the PR.

PR #1390, opened 2021-07-22
Skip custom JsonConverterAttribute when checking for JsonInheritanceConverter/StringEnumConverter

mrEDitor:
These methods currently throw `NullReferenceException` when tries to match `JsonConverterAttribute.ConverterType` but it's `null` (it's possible in `System.Text.Json`, see https://docs.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonconverterattribute.convertertype?view=net-5.0)


RicoSuter:
Thanks for the PR.

PR #1399, opened 2021-08-06
Support configuring whether constructor parameter are sorted

manfred-brands:
Replaces #1055


RicoSuter:
We should switch this to use `public required string Foo { get; init; }`, no?

PR #1403, opened 2021-08-16
Fixed validation of JTokens with type Uri

ertrumbl:
JsonSchemaValidator was not properly converting JTokens of type Uri to strings. Calling token.Value<string>() would throw exception because Uri does not implement IConvertible. Instead Uri's ToString() should be used to convert to string.

Converted to switch statement to handle multiple different Type scenarios and added unit test to cover Uri case.


ertrumbl:
@RicoSuter Is anyone still reviewing PR's that can take a look at this?

lahma (approved):
LGTM, thank you!

PR #1410, opened 2021-09-10
Fix liquid template to be inline with generate code

JohanAJ:
Fix KnockoutClass liquid template to be inline with generate code from property.ConvertToClassCode, and added assert to test method.


RicoSuter:
PR looks good.
Can you please fix the conflicts?

Issue #1267, opened 2020-10-13
why don't use StringComparison.OrdinalIgnoreCase on get property?


ehsanpishbaz:
hi, on json data usually ignore case sensitive keies, i expect use **obj?.Property(propertyInfo.Key, StringComparison.OrdinalIgnoreCase)** on [this](https://github.com/RicoSuter/NJsonSchema/blob/693d851d79ec466fa354f8a39521d98d8df3bd6c/src/NJsonSchema/Validation/JsonSchemaValidator.cs#L357) line.

RicoSuter:
I think properties in JSON Schema are not case insensitive, or are they? Can you post a link to the spec?

ehsanpishbaz:
it's true, JSON aren't case insensitive, but not bad if validation method get a parameter for do this.

lahma:
fixed by #1411

Issue #250, opened 2016-11-21
Ability to ignore case related errors.


holm0563:
I would like the ability to ignore errors regarding casing. 

The problem is validating json with improper casing ie {Name:"Derek"} vs {name:"Derek"} generates the error "NoAdditionalPropertiesAllowed". This is the same as the error I would get with the validation against {name1:"Derek"}.  What are the thoughts of adding a ValidationErrorKind.PropertyMismatch that makes it easier to distinguish this.

I'm happy to hope in and code this, just want to make sure its inline with the goals of this open source project first.

Thanks!

rsuter:
I don't think its a good idea to introduce PropertyMismatch because you cannot know whether its PropertyMismatch or NoAdditionalPropertiesAllowed...

2elsa:
@RicoSuter I'm interested in knowing if there is any timelines on when this will be added to the main branch. I currently have a requirement where I maintain a json schema in the API and client requests will be validated against this schema. Now, clients json could have different casing from the schema and `NoAdditionalPropertiesAllowed` doesn't look appropriate.

PR #1411, opened 2021-09-13
#250 Allow case insensitive property validation

asleire:
Added a `JsonSchemaValidatorOptions` class with a `IgnorePropertyNameCase` property as per your suggestion in #250

### DIFF
--- a/src/NJsonSchema/Validation/JsonSchemaValidatorOptions.cs
+++ b/src/NJsonSchema/Validation/JsonSchemaValidatorOptions.cs
@@ -3,7 +3,7 @@
 namespace NJsonSchema.Validation
 {
     /// <summary>Class to configure the behavior of <see cref="JsonSchemaValidator"/>. </summary>
-    public class JsonSchemaValidatorOptions
+    public class JsonSchemaValidatorSettings
     {
         /// <summary>Whether to ignore casing for object properties.</summary>
         public bool IgnorePropertyNameCase { get; set; }
--- a/src/NJsonSchema/Validation/JsonSchemaValidatorOptions.cs
+++ b/src/NJsonSchema/Validation/JsonSchemaValidatorOptions.cs
@@ -5,7 +5,13 @@ namespace NJsonSchema.Validation
     /// <summary>Class to configure the behavior of <see cref="JsonSchemaValidator"/>. </summary>
     public class JsonSchemaValidatorSettings
     {
-        /// <summary>Whether to ignore casing for object properties.</summary>
-        public bool IgnorePropertyNameCase { get; set; }
+        private StringComparer _propertyStringComparer;
+
+        /// <summary>The <see cref="StringComparer"/> used to compare object properties.</summary>
+        public StringComparer PropertyStringComparer
+        {
+            get => _propertyStringComparer ?? StringComparer.Ordinal;
+            set => _propertyStringComparer = value;
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


asleire:
Closes #1267

seungyongshim:
+1

src/NJsonSchema/Validation/JsonSchemaValidatorOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace NJsonSchema.Validation
+{
+    /// <summary>Class to configure the behavior of <see cref="JsonSchemaValidator"/>. </summary>
+    public class JsonSchemaValidatorOptions
+    {
+        /// <summary>Whether to ignore casing for object properties.</summary>
+        public bool IgnorePropertyNameCase { get; set; }
```
lahma, line 9: would it make sense to make this `public StringComparer PropertyNameComparer`?

src/NJsonSchema/Validation/JsonSchemaValidatorOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace NJsonSchema.Validation
+{
+    /// <summary>Class to configure the behavior of <see cref="JsonSchemaValidator"/>. </summary>
+    public class JsonSchemaValidatorOptions
```
RicoSuter, line 6: To be consistent, we should rename this to *Settings.

src/NJsonSchema/Validation/JsonSchemaValidatorOptions.cs
asleire: I don't mind either way. A `StringComparer` is obviously more flexible. @RicoSuter ?

RicoSuter: Fine for me either way.. as long it's an option.
To me it does not make much sense to allow case insensitive validation - it's probably against JSON Schema rules :-)

lahma: StringComparer has the advantage that you can write a comparer like `FirstCharIgnoringComparer` which is convenient in some interop scenarios between JSON and C# classes - not sure if it applies here.

asleire: I updated the PR with `StringComparer`

To me, I've always used case insensitive deserialization from JSON -> C# classes for convenience. Now I have a scenario where I want to use JsonSchema to validate certain JSON structures, and I want to preserve the case insensitivity for consistency

glecam:
First time using this library and it's great. I too would like to have this feature introduced for case insensitive deserialization from JSON -> C# classes.

RicoSuter:
Thanks for the PR.

asleire:
@Anit11 

Example usage:
 ```csharp
var jsonSchemaValidator = new JsonSchemaValidator(new JsonSchemaValidatorOptions()
{
    IgnorePropertyNameCase = true,
});

string json = ...; // Get this from somewehre
string jsonSchema = ...; // Get this from somewhere

var validationErrors = jsonSchemaValidator.Validate(json, jsonSchema);
```

> I want to use it for enum array property

Note that this PR is for property names, not property values. If you want case insensitive enum values that's not covered by this PR

Anit11:
Thanks for the example. I see, I actually want to validate the property value which is an enum array (case-insensitively). Is that supported? Below is the example and would like to validate Enum color (RED, BLUE)
public class ColorTypes
{
    public Color[] Fill { get; set; }
}
public enum Color {
    RED,
    BLUE
}
JsonSchema ColorSchema = JsonSchema.FromType<ColorTypes>();
ICollection<ValidationError> validationResult = ColorSchema .Validate(jsonObject);

PR #1412, opened 2021-09-14
TypeScript CodeGeneration checks AllowAdditionalProperties

cbalisky:
Code generation for TypeScript doesn't currently respect the AllowAdditionalProperties or AdditionalPropertiesSchema values. Current template code already has relevant syntax when HasIndexerProperty is set to true. HasIndexerProperty is only referenced by the Liquid template and is only used in ways that allows for additional properties in classes/interfaces, so I simply modified the getter for HasIndexerProperty to check the additional properties


pixellos:
@RicoSuter is that something that can be merged?

RicoSuter:
Thanks for the PR.

PR #1419, opened 2021-10-06
Enable support for C# 9.0 records

MaxWilson:
[Apologies in advance if I should have submitted an issue or something instead of just a PR out of the blue.]

This PR, in conjunction with a separate PR to NSwag, adds an option to change how ClassStyle="Record" generates DTOs. If you set **"classStyle": "Record"** and also **"UseNativeRecords": true**, nswag will generate DTOs as records instead of classes.

This allows you to use C# 9.0 "with" statements for usability, while still maintaining null safety.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/Models/ClassTemplateModel.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp/Models/ClassTemplateModel.cs
@@ -102,6 +102,9 @@ namespace NJsonSchema.CodeGeneration.CSharp.Models
         /// <summary>Gets a value indicating whether the class style is Record.</summary>
         public bool RenderRecord => _settings.ClassStyle == CSharpClassStyle.Record;
 
+        /// <summary>Gets a value indicating whether to generate records as C# 9.0 records.</summary>
+        public bool RenderAsNativeRecord => _settings.GenerateNativeRecords;
+
         /// <summary>Gets a value indicating whether to render ToJson() and FromJson() methods.</summary>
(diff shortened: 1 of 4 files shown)


MaxWilson:
I want to call out that in order to make the unit tests work, I had to upgrade the Microsoft.CodeAnalysis.CSharp dependency from 3.8.0 to 3.11.0, which is not compatible with .NET 4.5.2. I had to remove .NET 4.5.2 as a target framework for NJsonSchema.CodeGeneration.CSharp.Tests. I think this is a reasonable thing to do since it's test code, and doesn't impact NSwag/NJsonSchema's ability to generate .NET 4.5.2 code, but I want to make sure there's no pushback.

Alternately, we could just not AssertCompile() in that unit test, but I think that's a worse solution.

src/NJsonSchema.CodeGeneration.CSharp/Models/ClassTemplateModel.cs
```diff
@@ -102,6 +102,9 @@ public class ClassTemplateModel : ClassTemplateModelBase
         /// <summary>Gets a value indicating whether the class style is Record.</summary>
         public bool RenderRecord => _settings.ClassStyle == CSharpClassStyle.Record;
 
+        /// <summary>Gets a value indicating whether to generate records as C# 9.0 records.</summary>
+        public bool RenderAsNativeRecord => _settings.GenerateNativeRecords;
```
RicoSuter, line 106: I'd use GenerateNativeRecords as property name (to be aligned).

PR #1434, opened 2021-11-12
Use pattern variable to avoid several casts

testfirstcoder:
Use pattern variable for type check to avoid several casts.


RicoSuter:
Thanks for the PR.

PR #1435, opened 2021-11-15
Replace the zero-length array allocation with the method Array.Empty

testfirstcoder:
Avoid unnecessary memory allocation.


RicoSuter:
Thanks for the PR.

PR #1436, opened 2021-11-16
remove an unnecessary call ToList from a dictionary in a loop

testfirstcoder:


RicoSuter:
Are you sure it's not needed? 
I think this creates a copy so that the original list can be modified at the same time.

PR #1441, opened 2021-11-17
Add NUKE build and GitHub Actions

lahma:
I'm flying a bit blind here as this is the first commit, but the logic works locally, let's see.

* switching from .NET 5 to .NET 6 to makes things simpler (demos/tests)


lahma:
Oh man how slow Azure DevOps can be.

lahma:
@RicoSuter ready for review

lahma:
Closing in favor of #1446 which has local repo access.

PR #1446, opened 2021-11-22
Add NUKE build and GitHub Actions

lahma:
* switching from .NET 5 to .NET 6 to makes things simpler (demos/tests)


lahma:
@RicoSuter now the logic should be in place, but in order to do publish API keys are required (that is why build on windows-latest fails, it tries to publish), I have configured them as follows for Quartz.NET which has the same logic:

image

When secrets are in place, it should push to MyGet automatically (if on main, master or nuke branch and no tag present). 

When tag is pushed to master it should publish to NuGet, so a test tag could be `v11.0-beta-1` for example, this could then be unlisted from NuGet. These are steps that I'm afraid you need to test with your permissions and keys.

`Directory.Build.props` now contains all shared package information and copy-pasting is no longer needed. There's single `VersionPrefix` will be used with suffix. Suffix is date for MyGet Release (`NJsonSchema.CodeGeneration.CSharp.11.0.0-preview-20211122-1559.nupkg`) and for tag-based builds the tag is the version.

RicoSuter:
Secrets added... what about the failed actions?

RicoSuter:
Btw, I created discord servers for NSwag and NJsonSchema if you want to chat :-) /cc @jeremyVignelles

RicoSuter:
I see they are failing because of the myget publish... rerunning... but probably we only want to publish to myget on "master", not in PRs, no?

RicoSuter:
Ok, seems to be working as expected. Only strange thing: Why is it running the "build" (and MyGet push), shouldnt this run only on master/main/nuke and not on PRs?

lahma:
Sorry for the delay, I was out riding. This runs against master/main and *nuke* branch, we can remove the setup now as it's been verified (otherwise we would have needed to merge master to see it in action). I'll push a commit.

lahma:
> Btw, I created discord servers for NSwag and NJsonSchema if you want to chat :-)

I think it needs to be some sort of invite link, using the link https://discord.com/channels/912073155957374976/912073155957374980 directly gives:

image

Maybe test with incognito?

lahma:
I've cleaned the PR and ready from my part 🚀

jeremyVignelles:
I can confirm that the discord link doesn't work. It doesn't look like an invite link either (invite links are usually in the form `discord.gg/....`)

Right click on a channel, and click "Invite"

RicoSuter:
Changed the badges to invite links...

PR #1449, opened 2021-11-25
Add settings to generate only required fields with SampleJsonDataGenerator

steurer:
Currently, SampleJsonDataGenerator will generate all properties, regardless of whether they are required by the schema. Sometimes it can be useful to generate only the required fields, for example to generate minimum viable examples.

This PR adds settings that control the generation of required fields without changing the default behavior.

### DIFF
--- a/src/NJsonSchema.Tests/Generation/SampleJsonDataGeneratorTests.cs
+++ b/src/NJsonSchema.Tests/Generation/SampleJsonDataGeneratorTests.cs
@@ -134,7 +134,6 @@ namespace NJsonSchema.Tests.Generation
             Assert.Null(obj.Property("isoptional"));
         }
 
-
         [Fact]
         public async Task PropertyWithIntegerMinimumDefiniton()
         {
--- a/src/NJsonSchema.Tests/Generation/SampleJsonDataGeneratorTests.cs
+++ b/src/NJsonSchema.Tests/Generation/SampleJsonDataGeneratorTests.cs
@@ -89,6 +89,53 @@ namespace NJsonSchema.Tests.Generation
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


src/NJsonSchema/Generation/SampleJsonDataGeneratorSettings.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace NJsonSchema.Generation
+{
+    /// <summary> Settings for generating sample json data.</summary>
+    public class SampleJsonDataGeneratorSettings
+    {
+        /// <summary>Gets or sets a value indicating whether to generate to generate optional properties. (default: true).</summary>
```
RicoSuter, line 6: `Gets or sets a value indicating whether to generate optional properties (default: true).`

src/NJsonSchema.Tests/Generation/SampleJsonDataGeneratorTests.cs
```diff
@@ -88,6 +88,53 @@ public void Default_values_are_set_for_arrays()
...
+            //// Act
+            var token = generator.Generate(schema);
+            var obj = token as JObject;
+
+            //// Assert
+            Assert.NotNull(obj.Property("isrequired"));
+            Assert.Null(obj.Property("isoptional"));
+        }
+
+
```
RicoSuter, line 137: remove line

RicoSuter:
Thanks for the PR. 
However I currently cannot release a new version as there are already breaking changes in master and I'm working on v11.
Hopefully in the next weeks (there is a preview package however).

RicoSuter:
Is this ready?

steurer:
That is fine, I will use my fork in the meantime and switch to the official package once it's released.

Yes, from my side, this is ready. Thanks for the quick review

Issue #1447, opened 2021-11-23
... cannot have more than one member that has the attribute 'System.Text.Json.Serialization.JsonExtensionDataAttribute'.


jochenjonc:
I'm trying to use NSwag to generate a CSharp client (System.Text.Json) based on an OpenAPI spec from HERE. And with one of their APIs I have the following exception with the generated NSwag code.

https://developer.here.com/documentation/matrix-routing-api/8.5.0/swagger/spec.yaml

```
System.InvalidOperationException
The type 'Here.Routing.CalculateMatrix.Truck2' cannot have more than one member that has the attribute 'System.Text.Json.Serialization.JsonExtensionDataAttribute'.
   at System.Text.Json.ThrowHelper.ThrowInvalidOperationException_SerializationDuplicateTypeAttribute(Type classType, Type attribute)
   at System.Text.Json.Serialization.Metadata.JsonTypeInfo.CacheMember(Type declaringType, Type memberType, MemberInfo memberInfo, Boolean isVirtual, Nullable`1 typeNumberHandling, Boolean& propertyOrderSpecified, Dictionary`2& ignoredMembers)
```

The problem is that Truck2 inherits from Truck and both have the following code being generated.

``` csharp
private System.Collections.Generic.IDictionary<string, object> _additionalProperties = new System.Collections.Generic.Dictionary<string, object>();

[System.Text.Json.Serialization.JsonExtensionData]
public System.Collections.Generic.IDictionary<string, object> AdditionalProperties
{
	get { return _additionalProperties; }
	set { _additionalProperties = value; }
}
```

The base class and the derived class both have a property AdditionalProperties with the JsonExtensionData attribute and it seems that System.Text.Json (v6.0.0) doesn't allow that.

Is there a way to prevent AdditionalProperties being added to derived classes?
And what is the reason, the element in the yaml/json that renders this?

I want to help with a PR, but I'm a bit lost in the code ;-).

I found the place where the AdditionalProperties are generated.

https://github.com/RicoSuter/NJsonSchema/blob/463d246d0564c1d919503c15dbf0590b11ea0f1a/src/NJsonSchema.CodeGeneration.CSharp/Templates/Class.liquid#L117-L132

And the code that determines the HasAdditionalPropertiesType.
[...]

PR #1451, opened 2021-11-30
AdditionalProperties are only added on the base class

jochenjonc:
PR for #1447


RicoSuter:
I think this is not the right solution: 

- What if the base class does not allow additional properties only the subclass?
- What about a bigger inheritance tree?

RicoSuter:
Updated to use BaseClass instead, should cover these scenarios.

PR #1463, opened 2021-12-14
Fix ConvertCSharpDocs multi-line regression

### LINKED ISSUE #1462: Multiline comments generation issue
Hello, after PR #1346 code generation for multiline comments produces invalid `<summary>` xml doc.

For example:

```yaml
    TestType:
      type: object
      properties:
        code:
          type: string
          description: |
            * test1
            * test2
```

produces the following c# code:

```csharp
    [System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "13.15.3.0 (NJsonSchema v10.6.4.0 (Newtonsoft.Json v13.0.0.0))")]
    public partial class TestType
    {
        /// <summary>* test1
        /// <br/>* test2
    </summary>
        [Newtonsoft.Json.JsonProperty("code", Required = Newtonsoft.Json.Required.DisallowNull, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
        public string Code { get; set; }

        private System.Collections.Generic.IDictionary<string, object> _additionalProperties = new System.Collections.Generic.Dictionary<string, object>();

        [Newtonsoft.Json.JsonExtensionData]
        public System.Collections.Generic.IDictionary<string, object> AdditionalProperties
        {
            get { return _additionalProperties; }
            set { _additionalProperties = value; }
        }

    }
```

As you see the closing `</summary>` tag doesn't start with `///` which breaks the compilation

lahma:
Reverted back to old logic, added unit test, changed summary format to span multiple lines so multi-line ones don't look weird.

fixes #1462
fixes [#3793](https://github.com/RicoSuter/NSwag/issues/3793)

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp.Tests/GeneralGeneratorTests.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp.Tests/GeneralGeneratorTests.cs
@@ -293,7 +293,11 @@ namespace NJsonSchema.CodeGeneration.CSharp.Tests
             var output = generator.GenerateFile("MyClass");
 
             //// Assert
-            Assert.Contains(@"/// <summary>EnumDesc.</summary>", output);
+            var summary = @"
+        /// <summary>
+        /// EnumDesc.
+        /// </summary>".Replace("\r", "").Trim();
+            Assert.Contains(summary, output);
 
             AssertCompile(output);
         }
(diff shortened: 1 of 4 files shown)


RicoSuter:
It's a dangerous fix as it requires a template change (some might have custom templates).

lahma:
Old templates should work as-is. Now the built-in ones just add new line before and after the actual content:

```csharp
/// <summary>OLD_INSERTED_DATA</summary>
```

```csharp
/// <summary>
/// OLD_INSERTED_DATA
/// </summary>
```
The c# part is the sample, it just appends the same data to different valid doc comment section.

RicoSuter:
Need to fix those in NSwag as well, right?

image

lahma:
These should work without changes, but syncing the logic of having separate line might make sense.

PR #1473, opened 2022-01-04
Changed IsNullable for OpenApi3

Geronius:
Using NSwag to genereate OpenApi3 with 'openapi2csclient' the default was Mandatory for properties, this is not matching the OpenApi3 schema default, what has as default Optional. This resulted in not validated the jsonrequest-instances correctly.

Please merge this PR so I can use the openapi2csclient from you again instead of using my local one.


Geronius:
@RicoSuter, we still use a local version of this because of the change I made for this. Is there a reason not to change this in your version? 

I hope to hear from you.

PR #1476, opened 2022-01-10
Fix "Collection contains no matching element" error when no AllOf is of type "Object"

PatrickHofman:
This is a simple change to prevent the "Collection contains no matching element" InvalidOperationException that is thrown by `First()` when there is no AllOf of type "Object" (in this case, all are "None").


RicoSuter:
Thanks for the PR.

PR #1477, opened 2022-01-10
Fix "Collection contains no matching element" error when 'type' specifier is ommited in "allOf".

atmoos:
Bug happens when an allOf node contains two or more elements without a type specified. (apparently that's legal json schema)

Added a test to fix this issue from reoccurring. Given existing tests, this does not introduce a regression itself.


atmoos:
I see that PR #1476 fixes the same bug. I thought I'd leave this PR open in case the unit test included in this PR is of any value.

RicoSuter:
Thanks for the PR, Thomas! :-)

PR #1478, opened 2022-01-13
Support duration as of JSON Schema 2019-09

simonthum:
This introduces the "duration" format specifier as of JSON Schema 2019-09 as the preferred alternative to time-span. Following Postel's law, the change is intended to be compatible except for "duration" replacing time-span in schema output.


RicoSuter:
Does this also apply for OpenAPI 3?

simonthum:
As far as I can see, there is no OpenAPI 3 equivalent for duration, or another kind of time span, so it does not.

simonthum:
I adjusted the tests that failed due to the switch to duration.

simonthum:
Thank you!

PR #1481, opened 2022-01-16
Fixes #1480. Convert DateOnly properties to local timezone

### LINKED ISSUE #1480: Handle DateOnly in typescript client as local date (not UTC)
DateOnly is serialized to JSON as 'yyyy-MM-dd' (e.g. '2022-01-16').
When it's deserialized (in JS), NJsonSchema uses `new Date('2022-01-16')`.
Considering I'm in New York (UTC -5), `new Date('2022-01-16').toLocaleString()` gives me `1/15/2022, 7:00:00 PM`, and `new Date('2022-01-16').toLocaleDateString()` gives `1/15/2022`.
Moreover, if I now send this received DateOnly back to my API, it'd be sent as `2022-01-15`, because `formatDate` (used in NJsonSchema for `DateOnly` serialization) uses local date.

My personal expectation was to get it deserialized as `1/16/2022, 12:00:00 AM` (regardless of my time zone).

I understand, that JS by default interprets 'yyyy-MM-dd' as UTC date, but imo it'd be correct to adjust `DateOnly` fields after deserialization (or at least to make it configurable/provide some callbacks that I could set to adjust deserialization for DateOnly fields)

Shaddix:
Fixes #1480


src/NJsonSchema.CodeGeneration.TypeScript/Templates/ConvertToClass.liquid
```diff
@@ -11,8 +11,10 @@ if (Array.isArray({{ Value }})) {
 {%   if IsArrayItemNewableObject -%}
         {{ Variable }}{% if RequiresStrictPropertyInitialization %}!{% endif %}.push({{ ArrayItemType }}.fromJS(item{% if HandleReferences %}, _mappings{% endif %}));
 {%   else -%}
-{%     if IsArrayItemDate or IsArrayItemDateTime -%}
+{%     if IsArrayItemDate -%}
         {{ Variable }}{% if RequiresStrictPropertyInitialization %}!{% endif %}.push({{ StringToDateCode }}(item));
```
RicoSuter, line 15: Should be StringToDateOnlyCode and below StringToDateCode, no?

Shaddix: yes, you're right. sorry for mistake

Shaddix: resolved

RicoSuter:
If this PR should still be merged, please fix the conflicts and we'll have a look at it.

Shaddix:
conflicts are fixed

RicoSuter:
Thanks for the PR.

wuzzeb:
I think this PR doesn't completely fix the issue of dates and timezones, because serialization always assumes a local timezone.

Thus, if the setting ConvertDateToLocalTimezone is false, the parsing code will use new Date("2023-09-21") which assumes UTC but the formatting code calls formatDate (https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration.TypeScript/Templates/File.FormatDate.liquid) which uses `d.getMonth()` etc which assume local timezone.

This means if ConvertDateToLocalTimezone is false, you can parse and then immediately convert back to JSON and get a different result.

There should also be a `formatUTCDate` function which uses instead [getUTCMonth](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getUTCMonth) and friends, because dates are assumed to be in UTC.

Then, when ConvertDateToLocalTimezone is false, the generated `toJSON` function should use `formtUTCDate` instead of `formatDate`.

Then, both the serialization and deserialization will both use the same expected format and you can serialize followed by a deserialize without changing the data.

Issue #1128, opened 2020-02-05
Data not initialized in Constructor with Inheritance


tsinghammer:
We have the following structure generated in TypeScript from an Open API definition:

`export class ChildDto extends ParentDto implements IChildDto {
    someChildProp?: number | undefined;

    constructor(data?: IChildDto ) {
        super(data);
        this._discriminator = 'ChildDto ';
    }

    init(_data?: any) {
        super.init(_data);
        if (_data) {
            this.someChildProp= _data['someChildProp'];
        }
    }

    static fromJS(data: any): ChildDto {
        data = typeof data === 'object' ? data : {};
        let result = new ChildDto();
        result.init(data);
        return result;
    }

    toJSON(data?: any) {
        data = typeof data === 'object' ? data : {};
        data['someChildProp'] = this.someChildProp;
        super.toJSON(data);
        return data;
    }
}`

If i create a new instance of ChildDto by

`const additionalCashFlow = new IRSFixedCashFlowDto({ someParentProp: 100,  someChildProp: 1});`

the result is

`ChildDto{
        someParentProp: 100,
        someChildProp: undefined
        _discriminator: 'IRSFixedCashFlowDto',
      }`

where I would have expected someChildProp to be 1.

RicoSuter:
What is IRSFixedCashFlowDto?

tsinghammer:
Sorry, I forgot to replace this. It should be the ChildDto. I've edited the code example.

RicoSuter:
Ok, can you also post ParentDto?

tsinghammer:
Hi, yes. It looks like this: 

`export class ParentDto implements IParentDto {
    someParentProp?: number | undefined;

    protected _discriminator: string;

    constructor(data?: IParentDto) {
        if (data) {
            for (var property in data) {
                if (data.hasOwnProperty(property)) (<any>this)[property] = (<any>data)[property];
            }
        }
        this._discriminator = 'ParentDto';
    }

    init(_data?: any) {
        if (_data) {
            this.someParentProp = _data['someParentProp'];
        }
    }

    static fromJS(data: any): ParentDto {
        data = typeof data === 'object' ? data : {};
        if (data['t'] === 'SomeChildDto') {
            let result = new SomeChildDto();
            result.init(data);
            return result;
        }
        if (data['t'] === 'SomeOtherChildDto') {
            let result = new SomeOtherChildDto();
            result.init(data);
            return result;
        }
[...]

PR #1485, opened 2022-01-17
Fix for constructor inherited classes properties initialization when …

jposert:
…GenerateConstructorInterface & ConvertConstructorInterfaceData are true.

Simple template fix resolves problem reported in many places eg:
- https://github.com/RicoSuter/NJsonSchema/issues/1128
- https://github.com/RicoSuter/NSwag/issues/3580
- https://github.com/RicoSuter/NSwag/issues/2895
and probably few more.

The problem was when we tried to create instance of class which derives from another one.
After calling `super()` of the parent class, all of the descendant-specific properties are `undefined` and wont be initialized.
Problem example:

image

I changed Class template to include such initialization also in derived classes.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.TypeScript/Templates/Class.liquid
+++ b/src/NJsonSchema.CodeGeneration.TypeScript/Templates/Class.liquid
@@ -23,14 +23,12 @@
 {%     if HasInheritance -%}
         super({% if GenerateConstructorInterface %}data{% endif %});
 {%     endif -%}
-{%     if GenerateConstructorInterface and condition_temp -%}
         if (data) {
-{%         if HasInheritance == false -%}
             for (var property in data) {
-                if (data.hasOwnProperty(property))
+                if (data.hasOwnProperty(property) && this.hasOwnProperty(property))
                     (<any>this)[property] = (<any>data)[property];
             }
-{%         endif -%}
+{%     if GenerateConstructorInterface and condition_temp -%}
 {%         if ConvertConstructorInterfaceData -%}
 {%             for property in Properties -%}
 {%                 if property.SupportsConstructorConversion -%}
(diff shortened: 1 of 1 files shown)


RicoSuter:
Can you explain why it is undefined?
Wouldn't this produce even more code?

jposert:
Hi, at the moment I made this PR i wasn't really sure about the origin of the problem but now it's all clear for me.
So, it's common scenario when people use classes generated by NJsonSchema with Babel compiler (also in my case).

Babel compiles code that meets a standard https://tc39.es/proposal-class-fields/#sec-define-field
So as a result we've got javascript class like:
```javascript
class DeviceItem {
  constructor(data) {
    this.id = void 0;
    this.imageId = void 0;
    this.name = void 0;
    this._discriminator = void 0;

    if (data) {
      for (var property in data) {
        if (data.hasOwnProperty(property)) this[property] = data[property];
      }
    }

    this._discriminator = "DeviceItem";
  }
...
}
class DeviceBoardview extends DeviceItem {
  constructor(data) {
    super(data);
    this.components = void 0;
    this.tilesInfoUrl = void 0;
    this._discriminator = "DeviceBoardview";
  }
...
}
```
and here's the rub.
We assign all of the object properties in base's class constructor, no matter if it's property from derived or base class.
Unfortunately, because we don't assign the derived properties in the derived class declaration itself, and `super` must be a first statement in the constructor, we end with `undefined` properties.

So yeah, my solution produces like more code and also performs some duplicated assignments (because it assigns all the properties agin in derived class despite, some of the properties, that comes from the base class, are already correctly assigned). 
...but, it fixes the problem which many people experience.

**I think that best possible way to solve the problem is to just use `init` method in contructors.**

If you know any better solution let me know, im happy to implement it.

RicoSuter:
You cannot call init() because it accepts the "non-deserialized" original JSON structure whereas the ctor takes a 1:1 copy of the object - i.e. init transforms child-objects into class-object whereas the ctor value already has class-objects in properties...

To me it looks like Babel produces wrong JS, i.e. it initalizes them with undefined but as they have no assigned value they should not be touched at all. TS playground: 

image

PR #1489, opened 2022-02-02
Add support for TypeScript 4.3 override keyword

lahma:
Templates modified to emit override keyword when override supporting version detected (>= 4.3). Tested against our API generation and TS accepts when strict override mode is set.


ovska:
`4.3m` should probably be added also to GUI, `TypeScriptVersions` in `src/NSwagStudio/ViewModels/CodeGenerators/SwaggerToTypeScriptClientGeneratorViewModel.cs`

Issue #1486, opened 2022-01-25
Wrong DateFormatConverter used when UseSystemTextJson is specified


AroglDarthu:
I'm using NSwag toolchain v13.15.5.0 (NJsonSchema v10.6.6.0 (Newtonsoft.Json v11.0.0.0))
I've specified `/JsonLibrary:SystemTextJson`

A DTO uses a DateTimeOffset property:
``` CSharp
        [System.Text.Json.Serialization.JsonConverter(typeof(DateFormatConverter))]
        public System.DateTimeOffset StartDate { get; set; } = default!;
```

I'm seeing the following generated code:
``` CSharp
    [System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "13.15.5.0 (NJsonSchema v10.6.6.0 (Newtonsoft.Json v11.0.0.0))")]
    internal class DateFormatConverter : Newtonsoft.Json.Converters.IsoDateTimeConverter
    {
        public DateFormatConverter()
        {
            DateTimeFormat = "yyyy-MM-dd";
        }
    }
```

I would have expected (according to [src/NJsonSchema.CodeGeneration.CSharp/Templates/DateFormatConverter.liquid](https://github.com/RicoSuter/NJsonSchema/blob/265c5e9a43320cf86d74fde65370b8842dead6ac/src/NJsonSchema.CodeGeneration.CSharp/Templates/DateFormatConverter.liquid):
``` CSharp
internal class DateFormatConverter : JsonConverter<DateTime>
{
    public override System.DateTime Read(ref System.Text.Json.Utf8JsonReader reader, System.Type typeToConvert, System.Text.Json.JsonSerializerOptions options)
    {
        return DateTime.Parse(reader.GetString());
    }

    public override void Write(System.Text.Json.Utf8JsonReader writer, System.DateTime value, System.Text.Json.JsonSerializerOptions options)
    {
        writer.WriteStringValue(value.ToString("yyyy-MM-dd"));
    }
}
```

AroglDarthu:
@RicoSuter 
My guess is that the `DateFormatConverterTemplateModel` should either
- inherit `ClassTemplateModel` - or- 
- implement a property `UseSystemTextJson`

in order to properly access `UseSystemTextJson` value in liquid.

pekspro:
[...]

PR #1490, opened 2022-02-05
Fix DateFormatConverter for System.Text.Json

AroglDarthu:
Fix for #1486


RicoSuter:
Thanks for the PR. 
Question: Wouldnt it be better if it supports DateTime? or is the converter not called when a property is DT? and null in JSON?

AroglDarthu:
Contrary to how the converter is defined, `null` values get handled correctly. So, if your property is of type `DateTime?` the converter will correctly serialize and deserialize `null` values for it.
See https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?pivots=dotnet-5-0#handle-null-values

PR #1499, opened 2022-02-26
DateFormatConverter JsonException using full name

deathgore:
DateFormatConverter JsonException replaced with System.Text.Json.JsonException


RicoSuter:
Thanks for the PR.

PR #1502, opened 2022-03-10
Update JsonInheritanceConverter.liquid

zyofeng:
The JsonInheritanceConverters need to be made public so downstream NSwag Generator can get the DiscriminatorName (i.e. wrapping Client classes in another generated client class)

Error below:
6> ---> Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: 'System.Text.Json.Serialization.JsonConverter<BaseClass>' does not contain a definition for 'DiscriminatorName'


RicoSuter:
> The JsonInheritanceConverters need to be made public so downstream NSwag Generator can get the DiscriminatorName (i.e. wrapping Client classes in another generated client class)

I dont think that it needs to be public (as in reflection internal is the same as public I think)
Did you try the latest NSwag/NJS version as I think this has already been fixed.

zyofeng:
I'm using the latest NSwag.msbuild which is still affected. After manually modifying the accessor to public the problem went away.

zyofeng:
Alternatively we can put [assembly: InternalsVisibleTo("NJsonSchema")] 
This is probably a better solution.

zyofeng:
Also updated DateFormatConverter.liquid due to missing namespace.

RicoSuter:
Can you also fix the tests?

zyofeng:
Done.

Issue #1116, opened 2020-01-14
Wrong initializer generated


EricSch:
The generated code by NSwag.CodeGeneration.CSharp (Version="13.2.1") generates this code:
` public System.Collections.Generic.IDictionary<string, System.Collections.Generic.ICollection<string>> AssetDerivations { get; set; } = new System.Collections.Generic.Dictionary<string, System.Collections.ObjectModel.Collection<string>>();
`
But this code can't be compiled:
`Error	CS0266	Cannot implicitly convert type 'System.Collections.Generic.Dictionary<string, System.Collections.ObjectModel.Collection<string>>' to 'System.Collections.Generic.IDictionary<string, System.Collections.Generic.ICollection<string>>'. An explicit conversion exists (are you missing a cast?)`

The problem is between ObjectModel.Collection and Generic.ICollection. I know, I could change the configuration, but then I get similar problems in other places.

See also RicoSuter/NSwag#2496, seems to be a similar issue. Danke!

RicoSuter:
Moved to NJS.

quotschmacher:
Same here
`[Required(AllowEmptyStrings = false)]
[JsonProperty(Required = Required.DisallowNull, PropertyName = "DatabaseInsertRows")]
public List<SyncColumn[]> DatabaseInsertRows { get; set; }`

(SyncColumn is a simple object with two members)

generates 
`"databaseInsertRows": {
            "type": "array",
            "items": {
              "type": "array",
              "items": {
                "$ref": "#/components/schemas/SyncColumn"
              }
            }
          }`

in the swagger JSON

what leads to 
`public System.Collections.Generic.ICollection<System.Collections.Generic.ICollection<SyncColumn>> DatabaseInsertRows { get; set; } = new System.Collections.ObjectModel.Collection<System.Collections.ObjectModel.Collection<SyncColumn>>();`

and that cannot be compiled.

RicoSuter:
Ok so the bug here is that the inner collection must be the original interface type and not the instance type.

RicoSuter:
We'd need to fix this here somehow (it's a bit hacky right now - we need to refactor this quite a bit probably): 
https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration.CSharp/CSharpValueGenerator.cs#L60

aboryczko:
[...]

PR #1511, opened 2022-03-29
Fix for Wrong initializer generated #1116

aboryczko:
Fixes https://github.com/RicoSuter/NJsonSchema/issues/1116 and should also fix https://github.com/RicoSuter/NSwag/issues/3189


aboryczko:
Fixes #1116 and should also fix RicoSuter/NSwag#3189

PR #1513, opened 2022-04-15
Generate STJ JsonIgnoreAttributes for model properties

jcracknell:
You can't quite achieve the same behavior as Newtonsoft, as System.Text.Json doesn't deal in validation and JsonIgnoreAttribute only applies to the serialization process.


RicoSuter:
Serialization is more important than deserialization anyway. Would be nice to have both directions.
Thanks for the PR.

PR #1518, opened 2022-05-10
Change supported target frameworks

lahma:
Tp celebrate May 10 when also NET 5.0 has reached EOL, removing old targets that are hard to compile with stock Visual Studio & .NET SDK.

* change version to 12.0.0
* upgrade from net461 to net462
* drop support for net40, net45
* drop support for < netstandard2.0
* upgrade NUKE
* upgrade to Namotion.Reflection 3.0.0


RicoSuter:
Open task for v12: Update Namotion.Reflection

RicoSuter (approved):

lahma:
Namotion.Reflection upgraded to version 3.0.0 so I think now all the boxes have been ticked.

PR #1522, opened 2022-05-25
Fix output of JsonIgnore attributes

AroglDarthu:
Attempt to fix https://github.com/RicoSuter/NSwag/issues/4012
I'm not entirely certain on how to find out if the property is a value type.
Please have a look and tell me what you think.


src/NJsonSchema.CodeGeneration.CSharp/Models/PropertyModel.cs
```diff
@@ -64,17 +65,40 @@ public class PropertyModel : PropertyModelBase
...
+
+                bool IsValueType()
+                {
+                    var valueTypeFormats = new string[] { "date", "date-time", "uuid" };
+                    return _property.ActualTypeSchema.Type switch
+                    {
+                        JsonObjectType.Boolean => true,
+                        JsonObjectType.Integer => true,
+                        JsonObjectType.Number => true,
+                        JsonObjectType.String => valueTypeFormats.Contains(_property.Format),
```
lahma, line 95: this is a very slow way of checking, it uses LINQ extension against and does array looping, a better approach is just to do 

`_property.Format == "date" || _property.Format == "date-time || _property.Format == "uuid"`

AroglDarthu: Guess you're right.

I'm not sure if this is the way to determine value types though. If it is, we will need to determine the entire set of formats. For now I only added these three.

My concern is that `format` is an open value. See [Swagger data types format](https://swagger.io/docs/specification/data-models/data-types/#format)
So we might end up providing a setting for the generator. That setting will then obviously hold a collection of formats.

But perhaps there is a better way to determine if the property is a value type?

binginsin: Don't see it being that much slower since it will return once the item is found, and if it's not found in both cases it will go over all the values, whether looping or checking them via the OR operation. Besides this is not a time-critical application.

As for getting value types - if you check out https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration.CSharp/CSharpTypeResolver.cs  ResolveString() method, you can see which string formats NSwag supports currently, perhaps you could even use that method since it checks the Settings values for Date, DateTime and such. You already have an instance of type resolver in this class if you want to make use of it.

lahma: > Don't see it being that much slower since it will return once the item is found, and if it's not found in both cases it will go over all the values, whether looping or checking them via the OR operation. Besides this is not a time-critical application.

For single call, probably not that slow. It's O(n) now, but if called inside a loop to get all properties we enter O(n^2), add allocating array for every call, wrapping inside IEnumerable to enter LINQ, and then looping via boxed enumerator, it will be slow. We always need to consider how hot the code path is too. Suggested approach is O(1).

binginsin: > For single call, probably not that slow. It's O(n) now, but if called inside a loop to get all properties we enter O(n^2), add allocating array for every call, wrapping inside IEnumerable to enter LINQ, and then looping via boxed enumerator, it will be slow. We always need to consider how hot the code path is too. Suggested approach is O(1).

Wouldn't the valueTypeFormats array having a constant number of items make it O(1)?

lahma: > Wouldn't the valueTypeFormats array having a constant number of items make it O(1)?

No, loop is always O(n), n signifies how algorithm speed is affected by the (array) size.

binginsin: > No, loop is always O(n), n signifies how algorithm speed is affected by the (array) size.

This is probably not constructive for the review at this point, 
but in this case we would have n = length valueTypeFormats and m = number of properties. 
This particular check is O(n). 
This leaves us with a total complexity of O(n*m), however n is a constant (somewhere in the neighborhood of 10), which would leave us with O(10m) = O(m).
You suggested approach of combining values in valueTypeFormats with OR has exactly the same complexity, since each OR is essentially another if operation. So the complexity is bound by the number of OR statements we have, same as in the contains method.
There is definitely an overhead of wrapping to IEnumerable, but I don't think it's large enough to be considered a problem.

lahma: Ok I think I was taught this somewhat differently in CS classes in the university. Let's keep it that way and optimize later if needed. 

I did spent considerable time optimizing this library so that's my main reason being a bit hesitant for changes that can cause performance regressions.

AroglDarthu: @binginsin I like your suggestion. Would work like a charm if the type resolver returned a `System.Type`, but alas, it returns a string. Could still work if it would always be a fully qualified type name. Unfortunately it does not.

Most of the times it returns something like `Address` or `Person` or `string`. It als returns, for example `System.TimeSpan` and `System.TimeSpan?`.

So just using it and applying `System.Type.GetType(typeFromTypeResolver)?.IsValueType` is not a real option here :-/
Best option I can see now, is just to create a collection with the formats supported by the `CSharpTypeResolver` that we know are supposed to be value types.

Perhaps @RicoSuter can shine some light on this?

src/NJsonSchema.CodeGeneration.CSharp/Models/PropertyModel.cs
```diff
@@ -86,15 +93,18 @@ public string JsonIgnoreCondition
...
-                        JsonObjectType.Integer => true,
-                        JsonObjectType.Number => true,
-                        JsonObjectType.String => valueTypeFormats.Contains(_property.Format),
-                        _ => false,
+                        return true;
+                    }
+
+                    if (type.IsString())
+                    {
+                        return ValueTypeFormats.Contains(_property.Format);
```
AroglDarthu, line 104: Still not very happy with this. For instance, one could override the default type for `Date` in the _CSharpGeneratorSettings_. That means that a custom type could be used that is NOT a value type...

For now I really don't see a perfect solution to this problem. Other than maybe providing a _CSharpGeneratorSettings_ setting to specify the _ValueTypeFormats_ (which could default to the ones used now).

binginsin: I don't really see a good way around this either.. Adding another setting that will have to be maintained in the future is something @RicoSuter normally does not want to do.
Perhaps changing WhenWritingNull to WhenWiritngDefault would work better?

HavenDV:
I think you need to delete `ValueTypeFormats`

AroglDarthu:
> a look and tell me what you think.

binginsin:
@RicoSuter could you please have a look? This is still a serious issue and is causing problems, would be good to get this merged in.

RicoSuter:
Lgtm, thanks for the Pr.

abrasat:
@RicoSuter Could you please update the nuget packages with this fix?

bdovaz:
We need a NuGet package update, please!

PR #1525, opened 2022-05-30
Fixes #1524: typescript checking fails on Dictionary<Enum, string>

### LINKED ISSUE #1524: Typescript compilation fails for DTO with Dictionary<Enum, string> property
The following property in DTO:
```csharp
public Dictionary<DateTimeKind, string> EnumDictionary { get; set; }
```

Generates the following piece of code in `toJSON` method:
```typescript
        if (this.enumDictionary) {
            data["enumDictionary"] = {};
            for (let key in this.enumDictionary) {
                if (this.enumDictionary.hasOwnProperty(key))
                    (<any>data["enumDictionary"])[key] = this.enumDictionary[key];
            }
        }
```

Typescript complains:
```
error TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ readonly Unspecified?: string | undefined; readonly Utc?: string | undefined; readonly Local?: string | undefined; }'.
  No index signature with a parameter of type 'string' was found on type '{ readonly Unspecified?: string | undefined; readonly Utc?: string | undefined; readonly Local?: string | undefined; }'.

2217                     (<any>data["enumDictionary"])[key] = this.enumDictionary[key];
                                                              ~~~~~~~~~~~~~~~~~~~~~~~~
```
Apparantely, typescript assumes `key` is of type string.

Shaddix:


RicoSuter:
Thanks for the PR.

PR #1527, opened 2022-06-02
Fix DateFormatConverter System.Text.Json to respect DateType setting

### LINKED ISSUE #1526: CodeGeneration CSharp DateFormatConverter.liquid does not respect setting for DateType
When generating the DateFormatConverter, it now always uses `System.DateTime`. It should use the type specified for the _CSharpGeneratorSettings.DateType_ (which defaults to `System.DateTimeOffset`).

AroglDarthu:
Have `DateFormatConverterTemplateModel` pass on the `CSharpGeneratorSettings.DateType` setting, and use it in the liquid file. Fixes #1526


RicoSuter:
Thanks for the PR.

Issue #1519, opened 2022-05-20
Backwards compatability break 


drinehim:
Version 10.7.1 introduced a breaking change to the `JsonSchemaValidator` class in a patch (non-breaking) version update. It replaced the constructor `public JsonSchemaValidator(params IFormatValidator[] customValidators)` with `public JsonSchemaValidator(JsonSchemaValidatorSettings settings)`. This has broken a lot of our integrations with packages like Confluent.SchemaRegistry.Serdes.Json.

bruegth:
@RicoSuter @drinehim #1520 solves only constructor not the validate methods:

```
System.MissingMethodException: Method not found: 
System.Collections.Generic.ICollection`1<NJsonSchema.Validation.ValidationError>NJsonSchema.JsonSchema.Validate(System.String, NJsonSchema.Validation.FormatValidators.IFormatValidator[])'.
```

Before 10.7.1:
public ICollection<ValidationError> Validate(JToken token, params IFormatValidator[] customValidators);
After:
public ICollection<ValidationError> Validate(string jsonData, JsonSchemaValidatorSettings settings = null);


`Schema.Validate(strJson)` will fail at runtime if a newer package version is included somewhere else.

drinehim:
Thanks for pointing that out @bruegth just created PR #1557

RicoSuter:
Just ensure you use the same versions,
current releasable version will be v11 and it will anyway breaking.

PR #1557, opened 2022-09-27
Re-add old Validate functions to JsonSchema to fix backwards compatability issues.

drinehim:


drinehim:
Fix for #1519

RicoSuter:
Just ensure you use the same versions, 
current releasable version will be v11 and it will anyway breaking.

PR #1561, opened 2022-10-02
Recursive sample generation

### LINKED ISSUE #1560: Sample generation fails to generate values when re-using definitions in schema
Given this schema:
```json
{
    "definitions": {
        "withNumber": {
            "type": "object",
            "required": ["value"],
            "properties": {
                "value": {
                    "type": "number"
                }
            }
        }
    },
    "type": "object",
    "required": ["number1", "number2"],
    "properties": {
        "number1": { "$ref": "#/definitions/withNumber" },
        "number2": { "$ref": "#/definitions/withNumber" }
    }
}
```

I get the following JSON when using SampleJsonDataGenerator:
```json
{
  "number1": {
    "value": 0.0
  },
  "number2": null
}
```

Where I would expect it to generate the following:
```json
{
  "number1": {
    "value": 0.0
  },
  "number2": {
    "value": 0.0
  }
}
```

leflings:
Fixes #1560 

This PR modifies SampleJsonDataGenerator to support re-use of definitions, and by extension, recursion.

See issue for the problem before.

### Example 1

Now the following schema:
```json
{
    "definitions": {
        "withNumber": {
            "type": "object",
            "required": ["value"],
            "properties": {
                "value": {
                    "type": "number"
                }
            }
        }
    },
    "type": "object",
    "required": ["number1", "number2"],
    "properties": {
        "number1": { "$ref": "#/definitions/withNumber" },
        "number2": { "$ref": "#/definitions/withNumber" }
    }
}
```

generates 

```json
{
  "number1": {
    "value": 0.0
  },
  "number2": {
    "value": 0.0
  }
}
```

### Example 2

Schema:
```json
{
    "definitions": {
        "data": {
            "type": "object",
            "required": ["body"],
            "properties": {
                "body": { "$ref": "#/definitions/data" }
            }
        }
    },
    "type": "object",
    "required": ["data"],
    "properties": {
        "data": { "$ref": "#/definitions/data" },
    }
}
```

generates:

```json
{
  "data": {
    "body": {
      "body": {
        "body": null
      }
    }
  }
}
```

The level of recursion is controlled by a new setting on `SampleJsonDataGeneratorSettings`

### DIFF
--- a/src/NJsonSchema/Generation/SampleJsonDataGenerator.cs
+++ b/src/NJsonSchema/Generation/SampleJsonDataGenerator.cs
@@ -40,83 +40,95 @@ namespace NJsonSchema.Generation
         /// <returns>The JSON token.</returns>
         public JToken Generate(JsonSchema schema)
         {
-            return Generate(schema, new HashSet<JsonSchema>());
+            var stack = new Stack<JsonSchema>();
+            stack.Push(schema);
+            return Generate(schema, stack);
         }
 
-        private JToken Generate(JsonSchema schema, HashSet<JsonSchema> usedSchemas)
+        private JToken Generate(JsonSchema schema, Stack<JsonSchema> schemaStack)
         {
             var property = schema as JsonSchemaProperty;
             schema = schema.ActualSchema;
-            if (usedSchemas.Contains(schema))
+            try
             {
+                schemaStack.Push(schema);
+                if (schemaStack.Count(s => s == schema) > _settings.MaxRecursionLevel)
+                {
+                    return null;
+                }
+
+                if (schema.Type.IsObject() || GetPropertiesToGenerate(schema.AllOf).Any())
+                {
+                    var schemas = new[] { schema }.Concat(schema.AllOf.Select(x => x.ActualSchema));
+                    var properties = GetPropertiesToGenerate(schemas);
+
+                    var obj = new JObject();
+                    foreach (var p in properties)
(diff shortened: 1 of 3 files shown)


src/NJsonSchema/Generation/SampleJsonDataGenerator.cs
```diff
@@ -40,83 +40,95 @@ public SampleJsonDataGenerator(SampleJsonDataGeneratorSettings settings)
...
         {
             var property = schema as JsonSchemaProperty;
             schema = schema.ActualSchema;
-            if (usedSchemas.Contains(schema))
+            try
             {
-                return null;
-            }
+                schemaStack.Push(schema);
+                if (schemaStack.Count(s => s == schema) > _settings.MaxRecursionLevel)
```
lahma, line 55: doesn't this kill performance? capturing lambda and used inside recursion, having O(n) characteristics?

leflings: I would be very surprised if people use this functionality in performance sensitive scenarios. It's easy to rewrite this without the use of a capturing lambda if you deem this to be a big performance issue

lahma: Ah true, this is only for sample data indeed. Anything called by NSwag in normal generation process usually is quite performance sensitive.

PR #1563, opened 2022-10-08
Adds configuration settings to influence required annotation

tsinghammer:
Sometimes it would be useful, also with an OpenApi3 schema to generate non-required fields as Required.Default.
Do you think it makes sense to follow such a toggle and elaborate it more?


jkone27:
is this available for merge @RicoSuter ? this seettings seems useful instead of having to register a custom serializer to allow nullable properties (openapi schemas can often be unprecise in the real world)

PR #1567, opened 2022-10-21
Add settings UseUriTypeForUriFormattedStrings

herrahlstrom:
Now all uri formatted strings will translate into System.Uri, and it's not desired in every situation.
This change make it possible for the user to not use the System.Uri and string will be used instead.


RicoSuter:
Instead of adding more and more settings, I'd suggest to implement your own CSharpTypeResolver, override Resolve with this special rule and assign this to the settings...

PR #1570, opened 2022-10-24
Added abstract schema checking to CSharpValueGenerator.GetDefaultValue

gerg1:
Solution for issue: https://github.com/RicoSuter/NSwag/issues/4197


mathmul:
This'd be swell

PR #1578, opened 2022-12-03
Add API support for multiple schema extension definitions

### LINKED ISSUE #1568: Allow defining multiple keys
I have an attribute class that I want to use for defining multiple extended data keys, but the `IJsonSchemaExtensionDataAttribute` only supports a single key.

```
MyCustomAttribute : Attribute, IJsonSchemaExtensionDataAttribute
{
    public MyCustomAttribute(string test1value, string test2value) {
        Keys = new [] {"x-key1", "x-key2"}
        Values = new [] {test1value, test2value}
    }
}
```

The result is the schema generates 2 extended fields:
"x-key1": "test1value",
"x-key2": "test2value"

loligans:
This PR enables a developer to define attributes that allow multiple schema extensions to be defined.

Fixes #1568

### DIFF
--- a/src/NJsonSchema/Annotations/IMultiJsonSchemaExtensionDataAttribute.cs
+++ b/src/NJsonSchema/Annotations/IMultiJsonSchemaExtensionDataAttribute.cs
@@ -1,0 +1,12 @@
+using System;
+using System.Collections.Generic;
+
+namespace NJsonSchema.Annotations
+{
+    /// <summary>Interface to add multiple extension data property to a class or property, implementation needs to inherit from System.Attribute.</summary>
+    public interface IMultiJsonSchemaExtensionDataAttribute
+    {
+        /// <summary>Gets the extension data properties dictionary.</summary>
+        IDictionary<string, object> SchemaExtensionData { get; }
+    }
+}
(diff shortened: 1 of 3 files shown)


loligans:
Would appreciate a review whenever @lahma or @RicoSuter have the time. This isn't a big change but I'd be happy to hear any feedback you might have

src/NJsonSchema/Annotations/IMultiJsonSchemaExtensionDataAttribute.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using System.Collections.Generic;
+
+namespace NJsonSchema.Annotations
+{
+    /// <summary>Interface to add multiple extension data property to a class or property, implementation needs to inherit from System.Attribute.</summary>
+    public interface IMultiJsonSchemaExtensionDataAttribute
```
lahma, line 7: I'm not a subject matter expect here (Rico is), but would it make sore sense just to make a breaking change to change the support to have dictionary in original interface so there wouldn't be two ways to achieve the same thing (adding one item can be done via dictionary or via old interface).

loligans: this definitely could be done, I took the more conservative approach because I didn't want to break existing functionality. I don't mind refactoring to do what you suggest if that's what is agreed upon

loligans: would appreciate a review @RicoSuter when you have some time

RicoSuter:
v11

PR #1594, opened 2023-05-08
Don't use DateFormatConverter for DateOnly

Sinhk:
The DateFormatConverter has been causing some problems for use when using DateOnly.  
I don't think it's needed for DateOnly, and removing it will solve these problems


RicoSuter:
Seems to be a breaking change...

Issue #1588, opened 2023-01-31
StackOverflow occurs when you serialize inherited objects using STJ


shuaqq2004:
CODE：
```
[JsonInheritanceConverter(typeof(A), "$type")]
[JsonInheritanceAttribute("B", typeof(B))]
public class A
{
    public int? X { get; set; }
}

public class B : A
{
    public int? Y { get; set; }
}

A a = new A { X = 1 };
string s = JsonSerializer.Serialize(a); //error.
```
Code is generated via NSwagStudio, simplified DEMO.

![image](https://user-images.githubusercontent.com/3870105/215696766-2d42c2c1-0638-45b4-9401-1cfecd21aec5.png)

the suggestion in NET 7 , use `[JsonDerivedType]`

balchen:
I am experiencing this too.

The stack overflow happens when JsonInheritanceConverter<TBase>.Read is trying to determine the subtype to deserialize. From your example, this happens when you look at type A which has a JsonInheritanceConverter attribute. 

JsonInheritanceConverter<TBase>.Read calls JsonInheritanceConverter<TBase>.GetDiscriminatorType to get the subtype to use for deserialization. If the discriminator is not present in the serialized JSON stream, JsonInheritanceConverter<TBase>.GetDiscriminatorType will return the base type (A) again.

Deserializing A will result in JsonInheritanceConverter<TBase>.Read being called, and we have the endless loop causing the stack overflow.

It seems when a result type has subtypes with discriminator values, the discriminator must always be present, otherwise this endless loop will occur.

In my case, the return value is empty, but since the type is not nullable, it becomes an instance of A, serialized into JSON as {} in my case, with no discriminator, and so I see the same issue that you are seeing.

Perhaps that can help you solve your issue.

schnerring:
That's wrong:

```csharp
[JsonInheritanceConverter(typeof(A), "$type")]
```

It should be:

```csharp
[JsonInheritanceConverter(typeof(JsonInheritanceConverter), "$type")]
```

Note that `JsonInheritanceConverter` does not work with STJ yet.

PR #1595, opened 2023-05-12
Add support for STJ-native polymorphic `JsonDerivedType` and `JsonPolymorphic` attributes to C# client/schema generator

schnerring:
[Beginning with .NET 7, System.Text.Json supports polymorphic type hierarchy serialization and deserialization with attribute annotations.](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/polymorphism?pivots=dotnet-7-0). This PR adds support for the generation of `JsonPolymorphic` and `JsonDerivedType` attributes to the C# client generator.

This basically deprecates https://github.com/RicoSuter/NJsonSchema/blob/ed250a8b9c4f86a4da40cf0061cf0172cdea0449/src/NJsonSchema.Tests/Generation/SystemTextJson/JsonInheritanceConverter.cs#L14 for .NET 7 and later.

_It Works On My Machine™_, but the addition of unit tests and the changes I've made are pretty naive and the generated STJ code only works for .NET 7 or later. I was looking for something like a ".NET target type" parameter in the Liquid templates, but couldn't find any of the like. @RicoSuter I'd appreciate if you could point me to where else I gotta flesh out the code.

Related issues:

- https://github.com/RicoSuter/NJsonSchema/issues/1013
- https://github.com/RicoSuter/NJsonSchema/issues/1588
- https://github.com/RicoSuter/NSwag/issues/4375
- https://github.com/RicoSuter/NSwag/issues/3887
- https://github.com/RicoSuter/NSwag/issues/2243#issuecomment-780438323

### DIFF
--- a/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
+++ b/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
@@ -988,6 +988,26 @@ namespace NJsonSchema.Generation
 
             if (Settings.GenerateKnownTypes)
             {
+
+#if NET7_0_OR_GREATER
+                var jsonDerivedTypeAttributes = attributes
+                   // Known types of inherited classes will be generated later (in GenerateInheritance)
+                   .GetAssignableToTypeName(nameof(System.Text.Json.Serialization.JsonDerivedTypeAttribute), TypeNameStyle.Name)
+                   .OfType<Attribute>();
+
+                foreach (dynamic attribute in jsonDerivedTypeAttributes)
+                {
+                    if (attribute.DerivedType != null)
+                    {
+                        AddKnownType(attribute.DerivedType, schemaResolver);
+                    }
+                    else
+                    {
+                        throw new ArgumentException($"A KnownType attribute on {type.FullName} does not specify a type or a method name.", nameof(type));
+                    }
+                }
+#endif
+
                 var knownTypeAttributes = attributes
                    // Known types of inherited classes will be generated later (in GenerateInheritance)
                    .GetAssignableToTypeName("KnownTypeAttribute", TypeNameStyle.Name)
@@ -1184,7 +1204,13 @@ namespace NJsonSchema.Generation
         private object TryGetInheritanceDiscriminatorConverter(Type type)
         {
             var typeAttributes = type.GetTypeInfo().GetCustomAttributes(false).OfType<Attribute>();
-
+#if NET7_0_OR_GREATER
+            dynamic jsonPolymorphicAttribute = typeAttributes.FirstAssignableToTypeNameOrDefault(nameof(System.Text.Json.Serialization.JsonPolymorphicAttribute), TypeNameStyle.Name);
+            if (jsonPolymorphicAttribute != null)
+            {
+                return jsonPolymorphicAttribute;
+            }
+#endif
             dynamic jsonConverterAttribute = typeAttributes.FirstAssignableToTypeNameOrDefault(nameof(JsonConverterAttribute), TypeNameStyle.Name);
             if (jsonConverterAttribute != null)
             {
@@ -1207,6 +1233,12 @@ namespace NJsonSchema.Generation
 
         private string TryGetInheritanceDiscriminatorName(object jsonInheritanceConverter)
         {
+#if NET7_0_OR_GREATER
+            if (jsonInheritanceConverter is System.Text.Json.Serialization.JsonPolymorphicAttribute jsonPolymorphicAttribute)
+                return !string.IsNullOrWhiteSpace(jsonPolymorphicAttribute.TypeDiscriminatorPropertyName) ?
+                    jsonPolymorphicAttribute.TypeDiscriminatorPropertyName :
+                    JsonInheritanceConverter.DefaultDiscriminatorName;
+#endif
             return ObjectExtensions.TryGetPropertyValue(
                 jsonInheritanceConverter,
                 nameof(JsonInheritanceConverter.DiscriminatorName),
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "6.0.100",
+    "version": "7.0.100",
     "rollForward": "latestMinor"
   }
 }
--- a/src/NJsonSchema.CodeGeneration.CSharp.Tests/InheritanceInterfaceTests.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp.Tests/InheritanceInterfaceTests.cs
@@ -74,9 +74,9 @@ namespace NJsonSchema.CodeGeneration.Tests.CSharp
 
             //// Assert
             Assert.True(json.Properties["Item"].ActualTypeSchema.AllOf.First().HasReference);
-            Assert.Contains("[JsonInheritanceConverter(typeof(Fruit), \"discriminator\")]", code);
-            Assert.Contains("[JsonInheritanceAttribute(\"Banana\", typeof(Banana))]", code);
-            Assert.Contains("public class JsonInheritanceConverter<TBase> : System.Text.Json.Serialization.JsonConverter<TBase>", code);
+            Assert.Contains("[System.Text.Json.Serialization.JsonPolymorphic(TypeDiscriminatorPropertyName = \"discriminator\")]", code);
+            Assert.Contains("[System.Text.Json.Serialization.JsonDerivedType(typeof(Banana), typeDiscriminator: \"Banana\")]", code);
+            Assert.DoesNotContain("public class JsonInheritanceConverter<TBase> : System.Text.Json.Serialization.JsonConverter<TBase>", code);
         }
     }
 }
--- a/src/NJsonSchema.CodeGeneration.CSharp.Tests/InheritanceTests.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp.Tests/InheritanceTests.cs
@@ -135,14 +135,20 @@ namespace NJsonSchema.CodeGeneration.Tests.CSharp
         }
 
         [Fact]
-        public async Task When_definitions_inherit_from_root_schema()
+        public async Task When_definitions_inherit_from_root_schema_Newtonsoft()
         {
             //// Arrange
             var path = GetTestDirectory() + "/References/Animal.json";
 
             //// Act
             var schema = await JsonSchema.FromFileAsync(path);
-            var generator = new CSharpGenerator(schema, new CSharpGeneratorSettings { ClassStyle = CSharpClassStyle.Record });
+            var generator = new CSharpGenerator(
+                schema,
+                new CSharpGeneratorSettings
+                {
+                    ClassStyle = CSharpClassStyle.Record,
+                    JsonLibrary = CSharpJsonLibrary.NewtonsoftJson
+                });
 
             //// Act
             var code = generator.GenerateFile();
@@ -155,6 +161,33 @@ namespace NJsonSchema.CodeGeneration.Tests.CSharp
             Assert.Contains("[JsonInheritanceAttribute(\"PersianCat\", typeof(PersianCat))]", code);
         }
 
+        [Fact]
+        public async Task When_definitions_inherit_from_root_schema_STJ()
+        {
+            //// Arrange
+            var path = GetTestDirectory() + "/References/Animal.json";
+
+            //// Act
+            var schema = await JsonSchema.FromFileAsync(path);
+            var generator = new CSharpGenerator(
+                schema,
+                new CSharpGeneratorSettings
+                {
+                    ClassStyle = CSharpClassStyle.Record,
+                    JsonLibrary = CSharpJsonLibrary.SystemTextJson
+                });
+
+            //// Act
+            var code = generator.GenerateFile();
+
+            //// Assert
+            Assert.Contains("public abstract partial class Animal", code);
+            Assert.Contains("public partial class Cat : Animal", code);
+            Assert.Contains("public partial class PersianCat : Cat", code);
+            Assert.Contains("[System.Text.Json.Serialization.JsonDerivedType(typeof(Cat), typeDiscriminator: \"Cat\")]", code);
+            Assert.Contains("[System.Text.Json.Serialization.JsonDerivedType(typeof(PersianCat), typeDiscriminator: \"PersianCat\")]", code);
+        }
+
         private string GetTestDirectory()
         {
             var codeBase = Assembly.GetExecutingAssembly().CodeBase;
@@ -163,3 +196,4 @@ namespace NJsonSchema.CodeGeneration.Tests.CSharp
         }
     }
 }
+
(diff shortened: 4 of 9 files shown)


schnerring:
As an aside, it would be really, really nice to also generate Open API inheritance discriminators from the `[JsonPolymorphicAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonpolymorphicattribute)` instead of the `JsonConverterAttribute`. I think the magic happens here:

https://github.com/RicoSuter/NJsonSchema/blob/ed250a8b9c4f86a4da40cf0061cf0172cdea0449/src/NJsonSchema/Generation/JsonSchemaGenerator.cs#L1138-L1206

Do you think it's feasible to add support for this?

schnerring:
My use case is that I deserialize a document via [Marten](https://martendb.io/). The class used for deserialization is already annotated with the STJ polymorphism. I'd have to add `JsonConverter` and `KnownType` attributes just for NSwag.

```csharp
[JsonPolymorphic(TypeDiscriminatorPropertyName = "$t")]
[JsonDerivedType(typeof(FieldString), typeDiscriminator: nameof(FieldString))]
[JsonConverter(typeof(JsonInheritanceConverter), "$t")]
[KnownType(typeof(FieldString))]
public abstract class EntityFieldBase
{
    // ...
}
```

The attributes do conflict, so have to introduce a DTO class and AutoMapper as a workaround. As far as I can see,this would be resolved if the NSwag schema generator could also understand `JsonPolymorphic` and `JsonDerivedType` attributes.

Sorry for the noise, let me know if I should also open a corresponding issue for discussion.

schnerring:
> Do you think it's feasible to add support for this?

Turns out it wasn't too hard to make the C# generator aware of the new STJ polymorphic attributes. Have a look at the unit tests I've added to see how it works.

Todos:

- Add setting to `CSharpGeneratorSettings` so users are able to choose between STJ and "legacy" attributes.

src/NJsonSchema/Generation/JsonSchemaGenerator.cs
```diff
@@ -1184,7 +1204,13 @@ private void GenerateInheritanceDiscriminator(Type type, JsonSchema schema, Json
         private object TryGetInheritanceDiscriminatorConverter(Type type)
         {
             var typeAttributes = type.GetTypeInfo().GetCustomAttributes(false).OfType<Attribute>();
-
+#if NET7_0_OR_GREATER
+            dynamic jsonPolymorphicAttribute = typeAttributes.FirstAssignableToTypeNameOrDefault(nameof(System.Text.Json.Serialization.JsonPolymorphicAttribute), TypeNameStyle.Name);
+            if (jsonPolymorphicAttribute != null)
+            {
+                return jsonPolymorphicAttribute;
```
schnerring, line 1211: Not sure if it's a bit "too dirty" to use the JsonPolymorphicAttribute as inheritance converter.

cm-andrews:
It would be great if this could be merged. Nswag is literally the last piece of code we have to migrate before we can move to STJ.

franklixuefei:
Please merge! this is the last piece for us to get rid of our custom json converters and move completely to stj

franklixuefei:
@schnerring @RicoSuter My projects are under .NET 6 with STJ 7.0 installed. I verified that JsonPolymorphic works for .NET 6 projects. Should this PR also support NET 6?

src/NJsonSchema/Generation/JsonSchemaGenerator.cs
```diff
@@ -988,6 +988,26 @@ private void GenerateKnownTypes(Type type, JsonSchemaResolver schemaResolver)
 
             if (Settings.GenerateKnownTypes)
             {
+
+#if NET7_0_OR_GREATER
```
franklixuefei, line 992: should we also support NET 6? NET 6 with STJ 7.0 installed is also a valid combination, and I verified in my projects that `JsonPolymorphic` also works for STJ 7 under NET 6.

franklixuefei: Or maybe we could define a new constant `STJ7_0_OR_GREATER` use that as the condition

schnerring: > should we also support NET 6? NET 6 with STJ 7.0 installed is also a valid combination, and I verified in my projects that JsonPolymorphic also works for STJ 7 under NET 6.

Ah, you're right, the STJ 7.0.2 NuGet package supports `net6.0`. [It seems the docs aren't 100% accurate](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/polymorphism?pivots=dotnet-6-0): _In versions prior to .NET 7, System.Text.Json doesn't support the serialization of polymorphic type hierarchies._

> Or maybe we could define a new constant `STJ7_0_OR_GREATER` and use that as the condition

How would that work? Wouldn't the value of the directive depend on the .NET runtime version and also possibly an STJ NuGet dependency? I have no clue how to do that...

My idea was to focus on backwards compatibility. I'm not sure how common the usecase of .NET 6 + STJ 7 is. I can't say if it's worth the effort to support that scenario. If we want to support that, it's probably easier to just include the STJ 7 NuGet package for `net6.0`, `net462`, and `netstandard2.0`.

franklixuefei: >If we want to support that, it's probably easier to just include the STJ 7 NuGet package for net6.0, net462, and netstandard2.0.

hmmm I’m wondering… in this case, is this possible just to get rid of the NET7_0_OR_GREATER macro so that older .NET versions could leverage your work as well?

franklixuefei: I may need to take a closer look at the changes again, if getting rid of that macro is not possible/preferred. In general, I’m just trying to figure out a way to make your change available to net6 version or whichever version that STJ 7 supports.

franklixuefei: >Ah, you're right, the STJ 7.0.2 NuGet package supports net6.0. [It seems the docs aren't 100% accurate](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/polymorphism?pivots=dotnet-6-0): In versions prior to .NET 7, System.Text.Json doesn't support the serialization of polymorphic type hierarchies.

Yeah I agree… I’m thinking maybe what that doc wants to tell us is that for NET7 and above, STJ 7 is NATIVELY supported, which doesn’t necessarily mean that STJ 7 cannot be used in older NET versions.

franklixuefei:
This is still open? Please take a look @RicoSuter thanks!

jflygare:
A word of caution before migrating to STJ native polymorphism. STJ currently (.NET 7) only supports the discriminator as the first value found in the json object. There is an open request to address this.
https://github.com/dotnet/runtime/issues/72604
I ran into issues with this myself and had to implement my own converter anyway.

FWIW, here is my implementation of the STJ JsonInheritanceConverter/Attribute which addresses the StackOverflowException caused by recursive calls by the JsonSerializer.

  ```csharp
using System.Reflection;
using System.Text.Json;
using System.Text.Json.Serialization;
using System.Text.Json.Serialization.Metadata;

namespace .....
{

    /// <summary>
    /// <inheritdoc/>
    /// </summary>
    /// <typeparam name="TBase"></typeparam>
    public class JsonInheritanceConverter<TBase> : JsonConverter<TBase>
    {
        private readonly string _discriminatorName;
        private readonly Type _baseType;

        public JsonInheritanceConverter(string discriminatorName)
        {
            _discriminatorName = discriminatorName;
            _baseType = typeof(TBase);
        }

        /// <summary>
        /// <inheritdoc/>
        /// </summary>
        /// <param name="reader"></param>
        /// <param name="typeToConvert"></param>
        /// <param name="options"></param>
        /// <returns></returns>
        public override TBase Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
        {
            var document = JsonDocument.ParseValue(ref reader);
            var typeToReturn = document.RootElement.TryGetProperty(_discriminatorName, out var discriminator)
                ? GetObjectSubtype(typeToConvert, discriminator.GetString()!)
                : typeToConvert;

            // NOTE: To avoid calling this coverter recursively (StackOverflowException), we get the default converter for the base type
            //       and add to a "copy" of the input JsonSerializerOptions. This will take priority over the annotation and can be de-serialized
            //       as usual
            JsonSerializerOptions readOptions = options; 
            if (typeToReturn == _baseType)
            {
                var converter = (JsonConverter<TBase>)JsonTypeInfo.CreateJsonTypeInfo<TBase>(options).Converter;
                readOptions = new JsonSerializerOptions(options);
                readOptions.Converters.Add(converter);
            }
            return (TBase)JsonSerializer.Deserialize(document, typeToReturn, readOptions)!;
        }

        /// <summary>
        /// <inheritdoc/>
        /// </summary>
        /// <param name="writer"></param>
        /// <param name="value"></param>
        /// <param name="options"></param>
        public override void Write(Utf8JsonWriter writer, TBase? value, JsonSerializerOptions options)
        {
            Type typeToWrite = value!.GetType();
            var discriminatorValue = GetSubtypeDiscriminator(typeToWrite);

            // NOTE: To avoid calling this coverter recursively (StackOverflowException), we get the default converter for the base type
            //       and add to a "copy" of the input JsonSerializerOptions. This will take priority over the annotation and can be serialized
            //       as usual
            JsonSerializerOptions writeOptions = options;
            if (typeToWrite == _baseType)
            {
                var converter = (JsonConverter<TBase>)JsonTypeInfo.CreateJsonTypeInfo<TBase>(options).Converter;
                writeOptions = new JsonSerializerOptions(options);
                writeOptions.Converters.Add(converter);
            }

            
            var rootElement = JsonSerializer.SerializeToElement(value, typeToWrite, writeOptions);
            writer.WriteStartObject();
            // Look to see if discriminatorValue is not already a property and explicitly write if missing
            if (!rootElement.EnumerateObject().Any(e => e.Name.Equals(_discriminatorName, StringComparison.InvariantCultureIgnoreCase)))
                writer.WriteString(_discriminatorName, discriminatorValue);
            rootElement.EnumerateObject().ToList().ForEach(property => property.WriteTo(writer));
            writer.WriteEndObject();
        }


        private Type GetObjectSubtype(Type objectType, string discriminatorValue)
            => CustomAttributeExtensions.GetCustomAttributes<JsonInheritanceAttribute>(IntrospectionExtensions.GetTypeInfo(objectType), true)
                .Where(attrib => attrib.Key == discriminatorValue).Select(attrib => attrib.Type).FirstOrDefault() ?? objectType;


        private string GetSubtypeDiscriminator(Type objectType)
            => CustomAttributeExtensions.GetCustomAttributes<JsonInheritanceAttribute>(IntrospectionExtensions.GetTypeInfo(objectType), true)
                .Where(attrib => attrib.Type == objectType).Select(attrib => attrib.Key).FirstOrDefault() ?? objectType.Name;

    }

    /// <summary>
    /// <inheritdoc/>
    /// </summary>
    [AttributeUsage(AttributeTargets.Class)]
    internal class JsonInheritanceConverterAttribute : JsonConverterAttribute
    {
        public string DiscriminatorName { get; }

        public JsonInheritanceConverterAttribute(Type _, string discriminatorName = "discriminatorValue")
        {
            DiscriminatorName = discriminatorName;
        }

        public override JsonConverter? CreateConverter(Type baseType)
        {
            var converterType = typeof(JsonInheritanceConverter<>).MakeGenericType(baseType);
            return (JsonConverter?)Activator.CreateInstance(converterType, DiscriminatorName)
                ?? throw new InvalidOperationException($"Unable to create {converterType}");
        }
    }

}

schnerring:
See also: https://github.com/JasperFx/marten/issues/2586

Another _hack_ that works is choosing a discriminator property name like this:

```csharp
[JsonPolymorphic(TypeDiscriminatorPropertyName = "_")]
```

RicoSuter:
I think we need to put this behind an eg UseNativeInheritance flag (defalt false for now) so that you can opt-in to this new generator... if you look at the generated code as a black box then there should not really be a difference assuming both have the same features/performance...

schnerring:
I see your point, but wouldn't it be confusing for STJ users having to opt into this explicitly? Would it be reasonable to enable this feature by default if STJ is selected instead of Newtonsoft?

jeremia:
> I see your point, but wouldn't it be confusing for STJ users having to opt into this explicitly? Would it be reasonable to enable this feature by default if STJ is selected instead of Newtonsoft?

Since there is a possible complication with the issue https://github.com/dotnet/runtime/issues/72604 it wouldn't it be good to have the flag for now? The flag can be removed in the future when no such complications are known.

schnerring:
> The flag can be removed in the future when no such complications are known.

Fair enough.

If I understand correctly, some of the work from this PR has been integrated into the master branch (https://github.com/RicoSuter/NJsonSchema/commit/16bc1c038b899e41bba73bdd0061b85ff28af513).

I haven't kept up with the latest NSwag changes the past couple of months, so unless Rico or someone else tells me how to continue with the work of this PR, I won't add any changes.

jeremia:
Hi!
I was hoping that this fix would solve the issue https://github.com/RicoSuter/NSwag/issues/4375 thats bothering me.
So I was going to make a try to get this pull request finished.
It's my understanding that the actual code generation in NSwag studio is done by NJsonSchema in `NJsonSchema.CodeGeneration.CodeArtifact` in one of the constructors where the template's `Render` method is executed.
The call hierarchy goes here from both the code generation in NSwag Studio and the inheritance tests in NJsonSchema.

I cloned the schnerring branch of NJsonSchema and the main branch of NSwag.
I built the NJsonSchema and NJsonSchema.CodeGeneration NuGet packages with a new version number and put the packages in a local NuGet repo. I then updated the package references in NSwag for those packages and started NSwag Studio from Visual Studio. When I generated clients with .NET 7.0 using System.Text.Json with a schema containing inheritance the generated client did not contain the new native attributes `JsonDerivedType` or `JsonPolymorphic`, but still the old `JsonInheritanceConverter` and `JsonInheritanceAttribute`. It seems I'm missing something. Can anybody give me a clue?

It seems like NJsonSchema doesn't have a method for generating a client from a complete Open Api spec, but just the different parts. I was hoping to test the code generation of a complete example Open Api spec directly in NJsonSchema. Is that possible?

EelcoLos:
I'm wondering if this case is also covered in this PR: https://github.com/RicoSuter/NJsonSchema/issues/1666

schnerring:
I finally got the chance to take another look at this. The schema generation changes of this PR have already been incorporated in this commit: https://github.com/RicoSuter/NJsonSchema/commit/16bc1c038b899e41bba73bdd0061b85ff28af513

Since a lot of changes have been made since then, I created a new PR instead of rebasing this one. The new PR is my attempt to add the missing STJ polymorphism features to the C# client generator: https://github.com/RicoSuter/NJsonSchema/pull/1675

vvdb-architecture:
> A word of caution before migrating to STJ native polymorphism. STJ currently (.NET 7) only supports the discriminator as the first value found in the json object. There is an open request to address this. [dotnet/runtime#72604](https://github.com/dotnet/runtime/issues/72604) I ran into issues with this myself and had to implement my own converter anyway.

This has now been addressed with `AllowOutOfOrderMetadataProperties `

PR #1605, opened 2023-06-10
Allow setting x-csharp-existing-type on schemas

NaridaL:
to override default resolved type


NaridaL:
@RicoSuter any chance of getting this merged?

dthomas-inspire:
I think I also need this change or something similar.  To conform to the JSON Schema specification, I specify the "type" field of properties to be "integer", but I need the generated C# code to use various System types like `System.Int16` or `System.UInt32`, etc.  I don't see a way to do that without a change like this.

Issue #1618, opened 2023-07-26
TypeScript generation with EnumStyle `StringLiteral` and defaults


nycdotnet:
I am generating TypeScript code for an API with a MyClass schema defined like this:

```yaml
myProperty:
  type: "string"
  enum:
  - "ABC"
  - "DEF"
  - "GHI"
      default: "ABC"
```

I am generating with `/enumStyle:StringLiteral`.  This will correctly generate the type for myProperty which looks like this:

```typescript
export type MyClassMyProperty = "ABC" | "DEF" | "GHI";
```

However, when the default is used within the `init` method of `MyClass`, it still tries to use enum syntax which gives an error:

```typescript
this.myProperty= _data["myProperty"] !== undefined ? _data["myProperty"] : MyClassMyProperty.ABC;
```

I think this was just an oversight during the implementation of #1137 .  I am going to attempt to address this.

nycdotnet:
I am preparing a fix that overrides `GetEnumDefaultValue` in `TypeScriptValueGenerator` and changed the behavior if this is true:

```csharp
typeResolver is TypeScriptTypeResolver tsr && tsr.Settings.EnumStyle == TypeScriptEnumStyle.StringLiteral
```

nycdotnet:
Hooray!  I can confirm that upgrading `NSwag.MSBuild` to the latest preview (`14.0.0-preview008`) fixes our TypeScript generation issue!

PR #1619, opened 2023-07-27
Fix TypeScript enum default generation in StringLiteral mode

nycdotnet:
This fixes https://github.com/RicoSuter/NJsonSchema/issues/1618

I used the newer object pattern-matching syntax.  If this is a concern, I can do the check manually.


nycdotnet:
Will fix the raw string literals usage.

RicoSuter:
Thanks for the fix.
v11

PR #1621, opened 2023-08-14
enable decompression methods during http get operations

jamesmanning:
was trying to use nswagstudio with an openapi spec url of https://www.zuora.com/developer/yaml/swagger.yaml and it was failing since it's served with gzip compression and the automatic decompression isn't enabled (really wish that were the default behavior)


RicoSuter:
v11

PR #1622, opened 2023-08-17
Allign JsonPropertyRequiredCode with IsNullable

KristianHauge:
IsNullable implemented in 87bd7a4e3a552c5fab62564de598cb3ae6053d44

Currently the JsonPropertyRequiredCode does not reflect the previous fix, resulting in generating nullable properties with Required.DisallowNull and NullValueHandling.Ingore for optional properties.


RicoSuter:
v11

PR #1629, opened 2023-10-02
Optimize type resolving

lahma:
* reuse same `HashSet` for `reservedTypeNames` instead re-creating on every cache miss
* cleanup csprojs
* let NUKE regenerate GH actions definitions
* remove build configurations for `_build` project in solution file (prevents local command line usage)
* use string interpolation for better optimizations (NET 6+ has optimizations)

The test case has only 3 types and shows the difference quite nicely, real world usage can have hundreds of items, if not thousands.

## NJsonSchema.Benchmark.CSharpTypeResolverBenchmark

| **Diff**|Method|Mean|Error|Allocated|
|------- |-------|-------:|-------|-------:|
| Old |RegisterSchemaDefinitions|1.119 μs|0.0162 μs|1.66 KB|
| **New** |	| **645.8 ns (-44%)** | **5.40 ns** | **888 B (-48%)** |

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,20 +33,21 @@ jobs:
     name: windows-latest
     runs-on: windows-latest
     steps:
-      - uses: actions/checkout@v1
-      - name: Cache .nuke/temp, ~/.nuget/packages
-        uses: actions/cache@v2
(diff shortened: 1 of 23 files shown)


RicoSuter:
thanks

PR #1635, opened 2023-10-07
Add ability to customize which JsonProperties get ignored

DanielStout5:
I have a model where when serialized to JSON some properties should be ignored, but they should still be included in the JSON schema - I didn't see a way to do that in the library as-is. I put this version on NuGet here: https://www.nuget.org/packages/NJsonSchema-Ignore/


RicoSuter:
Can't you provide custom serializer options which exclude the property when serializing and include the property when generating the schema (ie two different options)?

PR #1639, opened 2023-10-26
Convert to use Central Package Management

lahma:
One file to rule them all.


lahma:
It should work. I use Rider where it works and they implemented after VS team did theirs. I don't have Visual Studio installation to test with though.

PR #1641, opened 2023-10-30
Extract annotations into own package

### LINKED ISSUE #1628: Create NJsonSchema.Attributes package
Hi,
are you open to extracting `NJsonSchema/Attributes` into a standalone package?

My motivation is that these attributes are used on the Domain/Contract level and most of your solution depends on these. That is fine by itself, but `NJsonSchema` has its dependencies like `Newtonsoft.Json` which we would like to push out of our codebase.

I know that `Newtonsoft` is heavily used in the library and I am not asking to remove it :) I am ok to use it in places where it makes sense.

If you are open to it, I can prepare a PR for it

trejjam:
Resolves #1628


trejjam:
In case you do not like `Resolve NUKE build deprecation` to be part of this PR I can move it to a separate PR

src/NJsonSchema.Annotations/JsonFormatStrings.cs
```diff
@@ -0,0 +1,84 @@
...
+// <license>https://github.com/RicoSuter/NJsonSchema/blob/master/LICENSE.md</license>
+// <author>Rico Suter, mail@rsuter.com</author>
+//-----------------------------------------------------------------------
+
+using System;
+
+namespace NJsonSchema.Annotations;
+
+/// <summary>Class containing the constants available as format string. </summary>
+public static class JsonFormatStrings
```
RicoSuter, line 14: I wonder whether we should put this in the NJsonSchema namespace but the NJsonSchema.Annotations pkg
and remove from NJsonSchema (and the same with JsonObjectType). What do you think?

trejjam: I understand the motive, but it feels strange to exclude one attribute from the annotations package.

The duplicity is there because of the Newtonsoft attribute. I can solve it instead with JsonConverter for that type. What do you think?

RicoSuter: Sorry no my propsal: 

JsonFormatStrings: Keep in NJsonSchema.Annotations but with namespace NJsonSchema (remove in NJsonSchema pkg)
JsonObjectType: Keep in NJsonSchema.Annotations but with namespace NJsonSchema (remove in NJsonSchema pkg)

trejjam: That's ok for me. How do you prefer to handle these [attributes](https://github.com/RicoSuter/NJsonSchema/pull/1641/files#diff-a21a5e443cda6224b4a0f1391771bfad4d1fd8f3b146a01879fa7fcf4a23e9fcL35)?

trejjam: They are from Newtonsoft package

RicoSuter: These? Link seems to be broken...

src/NJsonSchema.Annotations/NJsonSchema.Annotations.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
RicoSuter, line 4: we should probably go with

<TargetFrameworks>netstandard2.0;net462</TargetFrameworks>

src/NJsonSchema.Tests/Generation/AnnotationsGenerationTests.cs
```diff
@@ -15,10 +15,10 @@ public class AnnotationClass
         {
             public MyPoint Point { get; set; }
 
-            [JsonSchema(JsonObjectType.String, Format = "point")]
+            [JsonSchema(Annotations.JsonObjectType.String, Format = "point")]
```
RicoSuter, line 18: can you use "using NJsonSchema.Annotations;" here?

RicoSuter:
Ah ill update when it's merged.

src/NJsonSchema.Annotations/JsonFormatStrings.cs
trejjam: These: `JsonPropertyAttribute` 

https://github.com/RicoSuter/NJsonSchema/blob/8b30fcd633389a56c10ac23c9287348775f849a8/src/NJsonSchema/JsonObjectType.cs#L26C1-L28C19

```cs
using Newtonsoft.Json;

namespace NJsonSchema
{
    /// <summary>
    /// Enumeration of the possible object types.
    ///
    /// Keep in sync with <see cref="Annotations.JsonObjectType"/>
    /// </summary>
    [Flags]
    public enum JsonObjectType
    {
        /// <summary>No object type. </summary>
        [JsonProperty("none")] // <----------------------------- this ---------------<
        None = 0,

        /// <summary>An array. </summary>
        [JsonProperty("array")] // <----------------------------- and this ---------------<
        Array = 1,
...
    }
}
```

RicoSuter:
Sorry for merging, want to publish a new preview package ASAP.

src/NJsonSchema.Annotations/JsonFormatStrings.cs
RicoSuter: seems that these are not needed because of the magic here: 

image

RicoSuter:
v11.0.0-preview006

src/NJsonSchema.Annotations/JsonFormatStrings.cs
trejjam: Took me some time to find it :)

trejjam:
Sorry, I did not noticed unchecked checkbox:
image

PR #1642, opened 2023-10-31
Extract annotations into own package #1641 follow up

trejjam:
Move `JsonInheritanceAttribute` into Annotations package


trejjam:
If you do not find it useful, just close it

RicoSuter:
The problem is that in your model you’d also need the converter in the base class

RicoSuter:
Btw with net7 you dont need these anymore
https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/polymorphism?pivots=dotnet-6-0

trejjam:
I know that I personally do not need them. I just saw an attribute when I was searching about JsonConvertor<>. And this is the result.

I had in my mind that `NJS.NewtonsoftJson.Annotations` and `NJS.SystemTextJson.Annotations` might appear in the future. I do not think I need them now, but I am willing to prepare PR, so they will exist in the stable 11 release.

trejjam:
I did not know that `NewtonsoftJson` has its own `JsonInheritanceAttribute` in `NJsonSchema.NewtonsoftJson.Converters`. It makes this PR useless

RicoSuter:
The plan is that the default packages will eventually be stj only and newtonsoft lives in additional packges for legacy

PR #1646, opened 2023-11-08
Support configuring whether to generate StringEnumConverter

lahma:
STJ has problems with `JsonStringEnumConverter` when values come from `EnumMemberAttribute`. This can be bypassed by configuring serializer setting, but property setting win over them so there needs to be a switch to disable this generation. Defaults to true to work like before.


lahma:
The problem was that if there's attribute present, I cannot configure it more globally on serializer level. If there's attribute on property - it will win. So if I configure serializer to output X it won't happen when there's NJsonSchema generated property present that says it should be Y. The use case is to suppress the property generation and let the serializer use more global defaults.

Issue #1565, opened 2022-10-17
cannot load Swagger - The reader's MaxDepth of 64 has been exceeded 


marina-p:
When loading the Swagger document at https://github.com/Azure/azure-rest-api-specs/blob/main/specification/network/resource-manager/Microsoft.Network/stable/2022-05-01/networkInterface.json, the new default MaxDepth=64 in NewtonSoft.JSON (as of 13.0.1) is exceeded (call stack below).  An earlier version of this document could be loaded without issues by earlier versions of NSwag.

Is there any way currently to set the MaxDepth through the public interface when loading an OpenAPI document?  If not, would it be possible for you to expose this, or increase the maximum?

```
Unhandled exception. System.AggregateException: One or more errors occurred. (The reader's MaxDepth of 64 has been exceeded. Path 'definitions.NetworkInterfaceIPConfigurationPropertiesFormat.properties.applicationGatewayBackendAddressPools.items.properties.properties.properties.backendIPConfigurations.items.properties.properties.properties.virtualNetworkTaps.items.properties.properties.properties.destinationLoadBalancerFrontEndIPConfiguration.properties.properties.properties.publicIPAddress.properties.properties.properties.ipConfiguration.properties.properties.properties.subnet.properties.properties.properties.networkSecurityGroup.properties.properties.properties.networkInterfaces.items.properties.properties.properties.privateLinkService.properties.properties.properties.privateEndpointConnections.items.properties.properties.properties.privateEndpoint.properties.properties.properties.privateLinkServiceConnections.items.properties.properties.properties.privateLinkServiceConnectionState.properties.status', line 1, position 1306614.)
[...]

PR #1649, opened 2023-11-24
Increased Json MaxDepth to 128

johannesmols:
Using NSwag on schemas with large depths causes an exception to be thrown due to the MaxDepth default setting of 64. I am not aware of any way of customizing the Json settings for parsing the schema in a project when using NSwag, as it invokes the precompiled DLL directly. More context here: https://github.com/RicoSuter/NSwag/issues/4394. #1565 is also related, but can be solved by modifying the settings in the project.

I'm not sure if it is the best approach to just change the default setting in this project to solve it. There doesn't seem to be a way of passing in a `JsonSerializerSettings` object or func to modify it in `FromJsonAsync` (called from [here](https://github.com/RicoSuter/NSwag/blob/9175311f5427ccc899098971c8a4dd2353a1400a/src/NSwag.Core/OpenApiDocument.cs#L204C73-L204C73) in NSwag). If there was such an option, the default wouldn't have to be modified here but it could be changed in NSwag where the issue is relevant.

### DIFF
--- a/src/NJsonSchema/Infrastructure/JsonSchemaSerialization.cs
+++ b/src/NJsonSchema/Infrastructure/JsonSchemaSerialization.cs
@@ -214,7 +214,8 @@ namespace NJsonSchema.Infrastructure
                 MetadataPropertyHandling = MetadataPropertyHandling.Ignore,
                 ConstructorHandling = ConstructorHandling.Default,
                 ReferenceLoopHandling = ReferenceLoopHandling.Serialize,
-                PreserveReferencesHandling = PreserveReferencesHandling.None
+                PreserveReferencesHandling = PreserveReferencesHandling.None,
+                MaxDepth = 128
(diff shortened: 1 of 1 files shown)


johannesmols:
Apparently the unit tests are failing due to an old URL in a test, and is not related to this change. I made a new PR to fix this in #1651.

lahma (approved):
Seems like reasonable increment to me, consider it was unlimited before.

johannesmols:
Perfect. Perhaps it can then be released with the upcoming v11.

lahma:
We are waiting for Rico as he does the final call...

RicoSuter:
I think this is fine... we should not expose Newtonsoft.Json serializer internals as eventually NJS and NSwag will switch to STJ for schema serialization and this option would then go away anyway.

PR #1650, opened 2023-11-27
Add customizable number types in code generation

yolofy:
This commit introduces an update enabling the customization of number types during code generation.

This is particularly useful for consuming financial APIs where the usage of double and float may lead to the accumulation of rounding errors.

It introduces settings to customize the .NET type of number, float, double, and decimal data types during code generation in 'CSharpGeneratorSettings.cs' and updates 'CSharpTypeResolver.cs' to use these settings.

This change allows better user control over how the number types get translated, making the code generation more versatile. Unit tests for different scenarios are included in the new 'NumberTests.cs'.

Also fixed the url in When_yaml_OpenAPI_spec_is__served_with_gzip_compression__it_works from
- https://www.zuora.com/developer/yaml/swagger.yaml
to
+ https://developer.zuora.com/yaml/swagger.yaml


RicoSuter:
thx for the PR.

PR #1656, opened 2023-12-07
Adds CSharp System.Text.Json setting OptionalPropertiesWriteDefaultValue

### LINKED ISSUE #1564: Recent JsonIgnoreCondition change breaks NSwag C# clients
Caused by #1522

Example from string serialized enums, here is the source enum:
```cs
public enum Scope
{
    Value1 = 1,
    Value2 = 2,
    Value3 = 3,
}
```

Relevant part of OpenApi json:
```json
{
  // ...
  "Scope": {
    "enum": [
      "Value1",
      "Value2",
      "Value3"
    ],
    "type": "string",
    "x-enumNames": [
      "Value1",
      "Value2",
      "Value3"
    ]
  }
  // ...
}
```

Generated enum in C# client:
```cs
public enum Scope
{
    [System.Runtime.Serialization.EnumMember(Value = @"Value1")]
    Value1 = 0,
    [System.Runtime.Serialization.EnumMember(Value = @"Value2")]
    Value2 = 1,
    [System.Runtime.Serialization.EnumMember(Value = @"Value3")]
    Value3 = 2,
}
```

And finally the property w/ the enum in a generated class:
```cs
[System.Text.Json.Serialization.JsonIgnore(Condition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingDefault)]   
[System.Text.Json.Serialization.JsonConverter(typeof(System.Text.Json.Serialization.JsonStringEnumConverter))]
public Scope Scope { get; set; }
```

When `Value1` is used, the JSON property is never serialized, which causes the Scope enum to have an undefined value of `0` at the C# controller side.

The attribute can also have unintended consoquences when generating clients against runtimes that have different semantics from C#. For example, it's impossible to get the serializer to write a zero `int` or `DateTime.MinValue` property.
IMO the explicit ignore condition should never be used for value types, and I think it's debatable if it's required at all; when serializing you can opt to use `JsonSerializerOptions.DefaultIgnoreCondition`.

psiservices-dbrink:
fixes #1564

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp.Tests/JsonIgnoreAttributesTests.cs
+++ b/src/NJsonSchema.CodeGeneration.CSharp.Tests/JsonIgnoreAttributesTests.cs
@@ -173,5 +173,106 @@ namespace NJsonSchema.CodeGeneration.CSharp.Tests
                 Normalized(code)
             );
         }
+
+        [Fact]
+        public async Task When_using_SystemTextJson_and_OptionalPropertiesWriteDefaultValue_is_false_JsonIgnoreAttributes_are_still_generated_for_optional_properties()
+        {
+            //// Arrange
+            var schema = await JsonSchema.FromJsonAsync(@"{
+                ""type"": ""object"",
+                ""required"": [],
+                ""properties"": {
+                    ""optionalRef"": { ""type"": ""string"" },
+                    ""optionalValue"": { ""type"": ""integer"", ""format"": ""int32"" },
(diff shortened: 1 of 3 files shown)


RicoSuter:
I'd say instead of adding yet another option, we just revert all changes there: 
https://github.com/RicoSuter/NJsonSchema/issues/1564#issuecomment-1846820894

psiservices-dbrink:
@RicoSuter, the new beta package fixed my issue with enum values of 0.  Thank you.

PR #1662, opened 2024-01-10
Use generated property name for constructor parameters

fzzle:
Makes the naming of the constructor parameters consistent with the property names of the type when using the ParameterNameGenerator with records.


RicoSuter:
Thanks for the PR.

Issue #1305, opened 2021-01-17
CSharp Code Generation: GenerateOptionalPropertiesAsNullable is not respected for OAS Schema Components


Jesse0451:
Related to https://github.com/RicoSuter/NSwag/issues/2313

Optional Scalar properties do not produce nullable C# built-in value types when the types are defined in components/schema.

```yaml
openapi: '3.0.2'
info:
  title: API Title
  version: '1.0'
servers:
  - url: https://api.server.test/v1
paths:
  /test:
    post:
      requestBody:
        required: true
        content: 
         application/json:
            schema: 
              type: object
              properties:
                requiredInlineNumber:
                  type: integer
                  minimum: 1
                optionalInlineNumber:
                  type: integer
                  minimum: 1
                requiredSchemaNumber: 
                    $ref: "#/components/schemas/naturalNumber"
                optionalSchemaNumber: 
                    $ref: "#/components/schemas/naturalNumber"
              required: 
                - requiredInlineNumber
                - requiredSchemaNumber
              additionalProperties: false
      responses:
        '200':
          description: OK
components:
  schemas:      
    naturalNumber:
      type: integer
      minimum: 1
```
Generated C# class: 

```    [System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "10.3.1.0 (Newtonsoft.Json v11.0.0.0)")]
    public partial class Body 
    {
        [Newtonsoft.Json.JsonProperty("requiredInlineNumber", Required = Newtonsoft.Json.Required.Always)]
        [System.ComponentModel.DataAnnotations.Range(1, int.MaxValue)]
        public int RequiredInlineNumber { get; set; }
    
        [Newtonsoft.Json.JsonProperty("optionalInlineNumber", Required = Newtonsoft.Json.Required.DisallowNull, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
        [System.ComponentModel.DataAnnotations.Range(1, int.MaxValue)]
        public int? OptionalInlineNumber { get; set; }
    
        [Newtonsoft.Json.JsonProperty("requiredSchemaNumber", Required = Newtonsoft.Json.Required.Always)]
        [System.ComponentModel.DataAnnotations.Range(1, int.MaxValue)]
        public int RequiredSchemaNumber { get; set; }
    
[...]

PR #1682, opened 2024-03-14
Fix bug that C# property not nullable when specified via reference

markm77:
This fixes a bug that has been found using both NSwag and NJsonSchema. Essentially when `GenerateNullableReferenceTypes` and `GenerateOptionalPropertiesAsNullable` are both true, these settings are not respected for an optional object property provided via a reference (`$ref`) and the resulting C# property is not optional.

I believe this may solve multiple issues raised in both NSwag and NJsonSchema repos, for example:
https://github.com/RicoSuter/NJsonSchema/issues/1305
https://github.com/RicoSuter/NSwag/issues/4280 (maybe?)

I've added a test that reproduces the problem and passes with the fix.


markm77:
Hi @RicoSuter , is there anything I can do to get this fix into the repo? It's quite a pain to use a local version of NSwag every time to avoid this bug.

markm77:
Thanks for merging @RicoSuter !

PR #1697, opened 2024-04-28
Fix SystemTextJson read only properties handling

### LINKED ISSUE #1694: Private expression-bodied read-only properties should be ignored
Hi,

I'm currently encountering an issue with a generated OpenAPI schema that seems to be related on how properties are filtered by the `SystemTextJsonReflectionService` implementations.

```csharp
public class TestClass
{
    private string Prop1 { get; set; }

    private string Prop2 { get; } = "Hello";

    private string Prop3 => "Hello";
}
```

Both Prop2 and Prop3 are not filtered because they don't have a setter.

The excluding test, extracted from `SystemTextJsonReflectionService`, doesn't seem handle setter or getter nullability properly :
```csharp
if (accessorInfo.MemberInfo is PropertyInfo propertyInfo &&
    (propertyInfo.GetMethod?.IsPrivate == true || propertyInfo.GetMethod?.IsStatic == true) &&
    (propertyInfo.SetMethod?.IsPrivate == true || propertyInfo.SetMethod?.IsStatic == true) &&
    !propertyInfo.IsDefined(typeof(DataMemberAttribute)))
{
    continue;
}
```

Unlike the including test, from `NewtonsoftJsonReflectionService`

```csharp
contextualType
.Properties
.Where(p => p.PropertyInfo.DeclaringType == contextualType.Type &&
            (p.PropertyInfo.GetMethod?.IsPrivate != true && p.PropertyInfo.GetMethod?.IsStatic == false ||
             p.PropertyInfo.SetMethod?.IsPrivate != true && p.PropertyInfo.SetMethod?.IsStatic == false ||
             p.PropertyInfo.IsDefined(typeof(DataMemberAttribute))))
```

I think the `SystemTextJsonReflectionService` test should changed for:

```csharp
if (accessorInfo.MemberInfo is PropertyInfo propertyInfo &&
    (propertyInfo.GetMethod == null || propertyInfo.GetMethod.IsPrivate == true || propertyInfo.GetMethod.IsStatic == true) &&
    (propertyInfo.SetMethod == null || propertyInfo.SetMethod.IsPrivate == true || propertyInfo.SetMethod.IsStatic == true) &&
    !propertyInfo.IsDefined(typeof(DataMemberAttribute)))
{
    continue;
}
```

Regards

mereth:
The `SystemTextJsonReflectionService` didn't properly handle when a setter (or getter) was missing. As a result read-only private properties and read-only static properties were not ignored when building type schema via reflection.

Fixes #1694

### DIFF
--- a/src/NJsonSchema.Tests/Generation/SystemTextJson/SystemTextJsonTests.cs
+++ b/src/NJsonSchema.Tests/Generation/SystemTextJson/SystemTextJsonTests.cs
@@ -12,6 +12,14 @@ namespace NJsonSchema.Tests.Generation.SystemTextJson
             public string Name { get; }
 
             public string Description { get; }
+
+            private string PrivateReadOnlyProperty1 { get; }
+
+            private string PrivateReadOnlyProperty2 => "TEST";
+
+            public static string PublicReadOnlyStaticProperty { get; }
+
+            private static string PrivateReadOnlyStaticProperty { get; }
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


Erwinvandervalk:
I just ran into the same problem and wanted to raise the same PR as a fix :) Nice job! Hope it get's merged soon!

RicoSuter:
thanks for the PR.

PR #1701, opened 2024-06-07
Fixed SystemTextJson indexed properties handling.

viktoriia-zhel:
The `SystemTextJsonReflectionService` did not handle properly types with indexed properties. Exception "The JSON property 'item' is defined multiple times on type" was thrown if schema had been generated for type with multiple indexed properties.

In the applied fix indexed properties are detected and not added to the generated schema. 
Result schema matches schemas generated with `NewtonsoftJsonReflectionService` and previous versions of NJsonSchema library. 

Fixes https://github.com/RicoSuter/NSwag/issues/4874

### DIFF
--- a/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
+++ b/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
@@ -1089,7 +1089,7 @@ namespace NJsonSchema.Generation
             }
         }
 
-        private object? TryGetInheritanceDiscriminatorConverter(Type type)
+        private SystemTextJsonInheritanceWrapper? TryGetInheritanceDiscriminatorConverter(Type type)
         {
             var typeAttributes = type.GetTypeInfo().GetCustomAttributes(false).OfType<Attribute>();
 
--- a/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
+++ b/src/NJsonSchema/Generation/JsonSchemaGenerator.cs
@@ -1089,6 +1089,7 @@ namespace NJsonSchema.Generation
             }
         }
 
+#pragma warning disable CA1859
         private object? TryGetInheritanceDiscriminatorConverter(Type type)
         {
             var typeAttributes = type.GetTypeInfo().GetCustomAttributes(false).OfType<Attribute>();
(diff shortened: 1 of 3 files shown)


viktoriia-zhel:
@lahma Could you, please, help me with a build error that I get? 

I am getting " [ERR] Compile: D:\a\NJsonSchema\NJsonSchema\src\NJsonSchema\Generation\JsonSchemaGenerator.cs(1092,25): error CA1859: Change return type of method 'TryGetInheritanceDiscriminatorConverter' from 'object?' to 'NJsonSchema.Generation.JsonSchemaGenerator.SystemTextJsonInheritanceWrapper?' for improved performance (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1859) [D:\a\NJsonSchema\NJsonSchema\src\NJsonSchema\NJsonSchema.csproj::TargetFramework=netstandard2.0]". 

I tried to fix it as error message suggested, changed return type of method `TryGetInheritanceDiscriminatorConverter` to `SystemTextJsonInheritanceWrapper`. However, that broke unit test `When_schema_has_base_schema_then_it_is_referenced_with_Newtonsoft`. 

Do you know what is a correct type to use here to fix CA1859? 
Or can I disable this check for the method TryGetInheritanceDiscriminatorConverter so the build is successful?

src/NJsonSchema/Generation/JsonSchemaGenerator.cs
```diff
@@ -1089,7 +1089,7 @@ private void GenerateInheritanceDiscriminator(Type type, JsonSchema schema, Json
             }
         }
 
-        private object? TryGetInheritanceDiscriminatorConverter(Type type)
+        private SystemTextJsonInheritanceWrapper? TryGetInheritanceDiscriminatorConverter(Type type)
```
RicoSuter, line 1092: is this really needed? 
i think this might break CI?

viktoriia-zhel: Yes, I think this should be reverted, it breaks test `When_schema_has_base_schema_then_it_is_referenced_with_Newtonsoft`. 

However, without it build fails with " [ERR] Compile: D:\a\NJsonSchema\NJsonSchema\src\NJsonSchema\Generation\JsonSchemaGenerator.cs(1092,25): error CA1859: Change return type of method 'TryGetInheritanceDiscriminatorConverter' from 'object?' to 'NJsonSchema.Generation.JsonSchemaGenerator.SystemTextJsonInheritanceWrapper?' for improved performance (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1859) [D:\a\NJsonSchema\NJsonSchema\src\NJsonSchema\NJsonSchema.csproj::TargetFramework=netstandard2.0]"."

viktoriia-zhel: Can you suggest a better fix to CA1859?

lahma: You should be able to use 

```c#
#pragma warning disable CA1859
private object? TryGetInheritanceDiscriminatorConverter(Type type)
#pragma warning enable CA1859
```

Issue #1704, opened 2024-06-13
Annotations on separate lines?


martindamgaardlorensen:
In #1693, I introduced annotations for enums and enum members similar to how it was introduced to class and class properties in #516 and #522. I can however see that with the syntax used now:
> {%- template Xxx.Annotations -%}

..., any annotation, usually an attribute like:

> [Foo]
> [Bar(Baz = 42)]

... will be placed on the same line as the following declaration like:

> [Foo]
> [Bar(Baz = 42)]public class Bar
> { ... }

I would say the better option would be to have it as:

> {%- template Xxx.Annotations %}

... not stripping whitespace from the right side of the expansion.

However, it seems this has changed back and forth in the past, and I am unable to see any reasoning for this - so before making a PR removing the "-" on the right side in all 4 cases of template Xxx.Annotations, I would like to know if there are good reasoning behind having the dash there?

PR #1705, opened 2024-06-13
Allow attribute annotations to be on a separate line

martindamgaardlorensen:
See #1704


martindamgaardlorensen:
All the test that fails is a clear indication that this is not the right solution - and I am not qualified to find the right solution so I will close this PR.

Issue #1429, opened 2021-11-04
Could not find the JSON path of a referenced schema 


hyzx86:
It is recommended to throw a more detailed error message here
I debug the source code , here the clue is broken.

https://github.com/RicoSuter/NJsonSchema/blob/80c5d0014b2d5a3f5314af14742cb18ef94a5e53/src/NJsonSchema/JsonPathUtilities.cs#L71

How do I output the error information here?
``` C#
          if (mappings.Any(p => p.Value == null))
            {
                var errorItems = mappings.Where(p => p.Value == null).Select(x => x.Key.GetType().FullName);
                throw new InvalidOperationException("Could not find the JSON path of a referenced schema: " +
                                                     string.Join(",", errorItems) +
                                                    ",Manually referenced schemas must be added to the " +
                                                    "'Definitions' of a parent schema.");
            }
```

hyzx86:
996 what an amazing number 😂
Hoping to solve this problem, I went to try the following

JohnGalt1717:
I'm getting this error in nswag and I have no way to diagnose what the problem is as a result of this error message.

bencyoung-Fignum:
I'm also seeing this. When I look at which one is missing a value, it appears to be a JsonSchema object with no values at all, but I can't see where it's coming from:

![image](https://github.com/RicoSuter/NJsonSchema/assets/80325308/bd6cc263-0830-4f50-a0a3-daf2acc93d61)

![image](https://github.com/RicoSuter/NJsonSchema/assets/80325308/bcd11d31-f9b4-4a3d-8189-c0a0697cf067)

bencyoung-Fignum:
Looks like it's trying to resolve the schema for a JValue property but getting null as a reference and then throwing this error

hyzx86:
It might be possible to add some custom filters to locate the problem in the following way

```c#
            services.AddSwaggerGen(options =>
            {
                options.UseInlineDefinitionsForEnums();
                options.DocumentFilter<SwaggerDocumentFilter>();
                options.CustomSchemaIds(type => type.Name);
                options.SchemaFilter<SwaggerEnumSchemaFilter>();
                options.OperationFilter<SwaggerOperationIdFilter>();
                options.OperationFilter<SwaggerOperationFilter>();
                options.ParameterFilter<SwaggerEnumParameterFilter>();
                options.CustomDefaultSchemaIdSelector();
...
```

PR #1707, opened 2024-06-24
Throw more detailed error messages when encountering Json Schema errors

hyzx86:
fixes: #1429

### DIFF
--- a/src/NJsonSchema/JsonPathUtilities.cs
+++ b/src/NJsonSchema/JsonPathUtilities.cs
@@ -63,8 +63,10 @@ namespace NJsonSchema
 
             if (mappings.Any(p => p.Value == null))
             {
+                var errorItems = mappings.Where(p => p.Value == null).Select(x => x.Key.GetType().FullName);
                 throw new InvalidOperationException("Could not find the JSON path of a referenced schema: " +
-                                                    "Manually referenced schemas must be added to the " +
+                                                     string.Join(",", errorItems) +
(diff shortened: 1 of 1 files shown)


bencyoung-Fignum:
I don't think that will help me as there's no identifying names on the one that doesn't get a value. The only interesting thing about it is IsAnyType is true. However it has no properties or anything else naming it

image

src/NJsonSchema/JsonPathUtilities.cs
```diff
@@ -63,8 +63,10 @@ public static class JsonPathUtilities
 
             if (mappings.Any(p => p.Value == null))
             {
+                var errorItems = mappings.Where(p => p.Value == null).Select(x => x.Key.GetType().FullName);
                 throw new InvalidOperationException("Could not find the JSON path of a referenced schema: " +
-                                                    "Manually referenced schemas must be added to the " +
+                                                     string.Join(",", errorItems) +
+                                                    ",Manually referenced schemas must be added to the " +
```
RicoSuter, line 69:
```suggestion
                                                    ", Manually referenced schemas must be added to the " +
```

RicoSuter:
Is this still needed? I'm fine merging this as it's not really breaking or anything...

hyzx86:
> Maybe it produces way too much text in the exception message.

I think these texts are worthwhile. they can be somewhat easier for developers to locate the problem. 
After all, it has happened exceptionally 😅

RicoSuter:
Thanks for the PR.

PR #1711, opened 2024-06-30
feat: option to use c# 11 required keyword for required properties

alpoi-x:
This PR adds a `UseRequiredKeyword` option to the C# code generator. This will use the `required` [modifier](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/required), added in C#11, for any properties that are required in the json schema.


alpoi-x:
@RicoSuter ping!

alpoi-x:
@amis92 Isn't this already the case, given that a property is only treated as nullable if `GenerateOptionalPropertiesAsNullable` is enabled _and_ the property is not required?
https://github.com/RicoSuter/NJsonSchema/blob/9bf8f695b373410e8b51e1363270c08dda7b8127/src/NJsonSchema.CodeGeneration.CSharp/Models/PropertyModel.cs#L54-L55

amis92:
That, or the `null` is allowed by `type` JSON Schema keyword.

RicoSuter:
Lgtm, thanks for the PR.

Issue #1709, opened 2024-06-26
C# TimeSpan not resolved and generated as a DayJs duration


FreeFrags:
Im moving from MomentJS to DayJs in my code, i noticed that a C# TimeSpan was resolved as a moment,Duration. However now using DayJs it is resolving to a dayjs.Dayjs. 

I think this is the line we are talking about: 
https://github.com/RicoSuter/NJsonSchema/blob/1569e52401628d009af8044d1886d5a26770d05c/src/NJsonSchema.CodeGeneration.TypeScript/TypeScriptTypeResolver.cs#L290

My question is, is this a bug and should this line be dayjs.Duration or... is this done on purpose?

FreeFrags:
I created an issue in NSwag to, since if you decide this is an issue it might be nice to add the proper import too. 
https://github.com/RicoSuter/NSwag/issues/4925

PR #1714, opened 2024-07-03
#1709 Generate C# TimeSpan as a DayJs duration

FreeFrags:
Attempt to generate a Dayjs duration for a C# TimeSpan. 

See #1709

Im not familiar with the code however I wanted to attempt to help out. 

DayJS duration takes a ISO 8601 string or an object as a constructor. 

I chose to use a function to convert the 'D.HH:mm:ss.SSS' string to the object.

Im not sure if that is a good idea or if converting it to a ISO 8601 and passing that to the constructor would be more reusable.

Copilot suggests the following conversion function:
```
function convertToISO8601(input: string): string {
  const [days, rest] = input.split('.');
  const [hours, minutes, seconds, milliseconds] = rest.split(/[:.]/);

  // Constructing the ISO 8601 duration string
  let isoDuration = 'P';
  if (parseInt(days) > 0) isoDuration += `${parseInt(days)}D`;
  if (hours !== '00' || minutes !== '00' || seconds !== '00' || milliseconds !== '000') {
    isoDuration += 'T';
    if (parseInt(hours) > 0) isoDuration += `${parseInt(hours)}H`;
    if (parseInt(minutes) > 0) isoDuration += `${parseInt(minutes)}M`;
    if (parseInt(seconds) > 0 || parseInt(milliseconds) > 0) {
      let totalSeconds = parseInt(seconds);
      if (parseInt(milliseconds) > 0) {
        totalSeconds += parseInt(milliseconds) / 1000;
      }
      isoDuration += `${totalSeconds}S`;
    }
  }

  return isoDuration;
}
```
 
It would be best if the conversion function would be added as a function to the file as a utility which can then be called in the generated classes instead of inlined like i did here.

But im not sure how to do this and i hope someone can pick this up and polish it up.

### DIFF
--- a/src/NJsonSchema.CodeGeneration.TypeScript.Tests/DateCodeGenerationTests.cs
+++ b/src/NJsonSchema.CodeGeneration.TypeScript.Tests/DateCodeGenerationTests.cs
@@ -15,7 +15,7 @@ namespace NJsonSchema.CodeGeneration.TypeScript.Tests
 		'myTimeSpan': { 'type': 'string', 'format': 'time-span' }
 	}
 }";
-        
+
         [Theory]
         [InlineData(false)]
         [InlineData(true)]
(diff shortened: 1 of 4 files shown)


FreeFrags:
Please dont merge this. I will make a new pull request. 

I realized that the dayjs duration can also be constructed with a  ISO 8601.

PR #1716, opened 2024-07-05
C# generator: create anoymous types with camel casing

WolfgangHG:
This is an attempt to fix https://github.com/RicoSuter/NSwag/issues/4849 and https://github.com/RicoSuter/NSwag/issues/4837: since NSwag 14, the resulting C# file generated from a yaml file might contain lower case class names. 

I have no idea why this worked with NSwag 13 and what broke it.

Hope you don't force me to add a test case - for me it happens with a large closed source yaml file, and the open source file from the other bug report is also large. And I am not a yaml specialist...


WolfgangHG:
Just want to ping this...

lahma (approved):
Thank you.

PR #1726, opened 2024-08-29
Add C# Source Generator

valchetski:
## What's in this PR
This change aims to add a C# Source generator to create classes based on JSON schema. The idea is to make code generation more straightforward for generic scenarios. I've tried that approach on my project and it worked quite well. I thought it'd be good to make it part of the library. This Source Generator has few configurable properties and can be extended in the future if needed.
## Current approach
To generate C# classes you need to create a separate console project that will generate a `.cs` file and put it somewhere in your solution. Then you need to manually run that console application to generate/regenerate the models. Here's [example](https://github.com/bertt/geoparquet/blob/f4de4101e4b59a48bf3bbd3186d1126e000bec0c/geoparquet.codegen/Program.cs) 
## New (additional) approach
Add the following elements in the `.csproj` file of your project where the model will exist the following elements:
```
  <ItemGroup>
    <PackageReference Include="NJsonSchema.SourceGenerators.CSharp" />
  </ItemGroup>

    <ItemGroup>
        <AdditionalFiles 
            Include="schema.json" 
            NJsonSchema_GenerateOptionalPropertiesAsNullable="true" 
            NJsonSchema_Namespace="NJsonSchema.Demo.Generated" 
            NJsonSchema_TypeNameHint="PersonGenerated" 
            NJsonSchema_FileName="Person.g.cs" />
    </ItemGroup>
```
After that Source Generator will automatically create a `.cs` file during compilation. No additional console project or manual job is needed.
## Concerns
I have some questions/concerns regarding the implementation and put them in the comments. 
Also added some explanatory comments about why something was implemented that way. As tooling for the Source Generators is not the best for now there is a lot of msbuild magic in the `.csproj` files.


Directory.Packages.props
```diff
@@ -14,9 +14,12 @@
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="NodaTime" Version="3.1.9" />
     <PackageVersion Include="NSwag.Core.Yaml" Version="14.0.0-preview011" />
+    <PackageVersion Include="Parlot" Version="0.0.25.0" />
```
valchetski, line 17: Source generators do not support transient dependencies: https://github.com/dotnet/sdk/issues/17775
Because of that need to explicitly install some transient packages.
For example, it'll be used here to include `Parlot` into NuGet package https://github.com/RicoSuter/NJsonSchema/pull/1726/files#diff-ead3877feb09b2619c61d0feb8b85e0c85c627756febe0dda0829dafe13d5325R39

src/NJsonSchema.SourceGenerators.CSharp/NJsonSchema.SourceGenerators.CSharp.csproj
```diff
@@ -0,0 +1,74 @@
...
+    <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
+    <LangVersion>latest</LangVersion>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <IsRoslynComponent>true</IsRoslynComponent>
+    <NoWarn>1701;1702;NU5128</NoWarn>
+  </PropertyGroup>
+    
+  <ItemGroup>
+    <ProjectReference Include="..\NJsonSchema.CodeGeneration.CSharp\NJsonSchema.CodeGeneration.CSharp.csproj" PrivateAssets="All">
```
valchetski, line 15: With `PrivateAssets="All"` it will not add `NJsonSchema.CodeGeneration.CSharp` as a dependency of `NJsonSchema.SourceGenerators.CSharp` project as I assume it's not needed. Code from `NJsonSchema.SourceGenerators.CSharp` should be enough to generate C# classes. For more advanced scenarios `NJsonSchema.CodeGeneration.CSharp` can be installed explicitly in client's projects.

src/NJsonSchema.SourceGenerators.CSharp/NJsonSchema.SourceGenerators.CSharp.csproj
```diff
@@ -0,0 +1,74 @@
...
+        <SetTargetFramework>TargetFramework=netstandard2.0</SetTargetFramework>
+    </ProjectReference>
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" PrivateAssets="all" />
+    <PackageReference Include="Newtonsoft.Json" GeneratePathProperty="true" />
```
valchetski, line 23: Deliberately didn't add `PrivateAssets="all"` for `Newtonsoft.Json` so it will be included as NuGet dependency. In the generated code Newtonsoft attributes will be used. That way client don't need to install `Newtonsoft.Json` in their project explicitly.

src/NJsonSchema.SourceGenerators.CSharp/NJsonSchema.SourceGenerators.CSharp.csproj
```diff
@@ -0,0 +1,74 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
valchetski, line 1: My main concern is whether we should create a separate project for the source generator or add them to the existing `NJsonSchema.CodeGeneration.CSharp`. Below are my thoughts.
## Add Source Generator to new `NJsonSchema.SourceGenerators.CSharp`
### Pros:

- Clients will install that package only if they know they need to use a source generator.
- Easier to maintain as source generator projects require additional configuration (`IsRoslynComponent`, exclude NU5128 warning, etc.)
### Cons:
- The names of the NuGet packages could be confusing. `NJsonSchema.CodeGeneration.CSharp` and `NJsonSchema.SourceGenerators.CSharp` may sound a bit synonimical.
## Add Source Generator to existing `NJsonSchema.CodeGeneration.CSharp`
### Pros:
- We don't introduce a new NuGet package.
- By installing `NJsonSchema.CodeGeneration.CSharp` client can use either classes from the library or/and Source Generator.
### Cons: 
- It'll automatically install the analyzer when the client does a package upgrade. (However, the analyzer will not do much without additional configuration. Not sure if this is an issue).
- The project might be harder to maintain as it will be a mix of regular code and code that will go to the source generator. (Frankly speaking, I didn't try that approach at all, but in theory it should work).

HavenDV:
You can check my implementation of Source Generator for NSwag to see how to support transient packages automatically https://github.com/HavenDV/H.NSwag.Generator

valchetski:
> You can check my implementation of Source Generator for NSwag to see how to support transient packages automatically https://github.com/HavenDV/H.NSwag.Generator

Thank you @HavenDV. Amazing library 🙂 However, I have some issues with project references. I asked about them here: HavenDV/H.Generators.Extensions#15

seesharprun:
@RicoSuter

PR #1729, opened 2024-09-07
fixes #1728 return null when subType is not found, fixes Stackoverflow exception

### LINKED ISSUE #1728: Unknown subtype causes Stackoverflow exception at deserialization
See example code & payload

```
[JsonInheritanceConverter(typeof(Event), "$type")]
[JsonInheritanceAttribute("Arrival", typeof(Arrival))]
[JsonInheritanceAttribute("Departure", typeof(Departure))]
public partial class Event
{
}

public partial class Data
{
    [JsonPropertyName("events")]
    public ICollection<Event> Events { get; set; } = new System.Collections.ObjectModel.Collection<Event>();
}

{ "events": [{"$type":"Cancel"}] }
```

trying to deserialize the payload with a (yet) unknown subtype "Cancel" causes a stackoverflow exception.

I think the problem is with the generated ```JsonInheritanceConverter<TBase>.GetObjectSubtype```.
It always returns a value but it should return ```null``` when it cannot find any known subtype.

```
private System.Type GetObjectSubtype(System.Type objectType, string discriminator)
{
    foreach (var attribute in System.Reflection.CustomAttributeExtensions.GetCustomAttributes<JsonInheritanceAttribute>(System.Reflection.IntrospectionExtensions.GetTypeInfo(objectType), true))
    {
        if (attribute.Key == discriminator)
            return attribute.Type;
    }

    return objectType;  // <--- this should return null!
}
```

The only place where ```GetObjectSubtype``` is called is in ```JsonInheritanceConverter<TBase>.GetDiscriminatorType``` which tests for null(!) ...

When the fix (returning null when subtype is not found) is applied the code throws the correct exception
```
throw new System.InvalidOperationException("Could not find subtype of '" + objectType.Name + "' with discriminator '" + discriminatorValue + "'.");
```

ryanheath:
Fixes #1728 

I do not know how to test it. The generated code in the liquid file is not the same a the 'real' ```NJsonSchema.Converters.JsonInheritanceConverter<TBase>```, which has tests.

Also the 'real' code of ```GetObjectSubtype``` returns a null value when subType is not found.
so the fix got to be right! ;)

```
private static Type? GetObjectSubtype(Type baseType, string discriminatorValue)
{
    var jsonInheritanceAttributes = baseType
        .GetTypeInfo()
        .GetCustomAttributes(true)
        .OfType<JsonInheritanceAttribute>();

    return jsonInheritanceAttributes.SingleOrDefault(a => a.Key == discriminatorValue)?.Type;
}
```

### DIFF
--- a/src/NJsonSchema.CodeGeneration.CSharp/Templates/JsonInheritanceConverter.liquid
+++ b/src/NJsonSchema.CodeGeneration.CSharp/Templates/JsonInheritanceConverter.liquid
@@ -100,7 +100,7 @@ public class JsonInheritanceConverter<TBase> : System.Text.Json.Serialization.Js
                 return attribute.Type;
         }
 
-        return objectType;
+        return null;
     }
 
     private string GetSubtypeDiscriminator(System.Type objectType)


RicoSuter:
Fixes https://github.com/RicoSuter/NSwag/issues/4375

RicoSuter:
Thanks for the PR.

PR #1730, opened 2024-09-09
Bump Fluid.Core

rexcfnghk:
[Fluid.Core](https://github.com/sebastienros/fluid/blob/0de3b07a5eeba650d69f19849fa62255fa1b4ca9/Fluid/FluidParser.cs#L522) has a breaking change between 2.9.0 and 2.11.1 because Parlot, the library Fluid.Core depends on, has a [breaking change](https://github.com/sebastienros/parlot/pull/103/files#diff-075f3dea19a91e058660dfe2a77be11a72a2d4fbbcf24441ec6a5f9a46869540L9) which propagated to Fluid.Core between 2.9.0 and 2.11.1.

This change bumps Fluid.Core and transitively upgraded Parlot to bring in the change.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -5,7 +5,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageVersion Include="BenchmarkDotNet" Version="0.13.9" />
-    <PackageVersion Include="Fluid.Core" Version="2.9.0" />
+    <PackageVersion Include="Fluid.Core" Version="2.11.1" />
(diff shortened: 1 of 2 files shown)


lahma:
There's already #1719

rexcfnghk:
Would it better to separate #1719 and this PR? To me #1719 seems more like a housekeeping PR but this PR fixes an actual issue due to transitive dependencies

RicoSuter:
Done in other PR. Still thanks for your PR.

PR #1733, opened 2024-09-26
configurable fieldname

notlazychen:
Because the naming convention for properties in JsonSchema is not very standardized, we have property names like "_id".
When the CSharpGeneratorSettings.ClassStyle setting is Inpc, the field name generation logic causes the automatically generated field _id for the property Id to have the same name as the property _id. We hope to be able to customize the prefix.


RicoSuter:
Thanks for the PR.

PR #1743, opened 2024-11-19
Prepare for NET 9 SDK analysis

lahma:
NET 9 SDK has more rules and will cause build problems. Creating a TODO list for focused PRs in `Directory.Build.props` via `NoWarn` items, fixing small easy wins inside the PR. Now compiles with NET 9 SDK. All changes made with IDE automation.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,6 +48,7 @@ jobs:
           MYGET_API_KEY: ${{ secrets.MYGET_API_KEY }}
       - name: 'Publish: artifacts'
         uses: actions/upload-artifact@v4
+        if: runner.os == 'Windows'
         with:
           name: artifacts
           path: artifacts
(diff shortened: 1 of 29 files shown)


RicoSuter:
@lahma can you maybe also add a fix for this windows CI build failure (conflict)?

image

lahma:
Sure, gimme an hour so that I get back to computer, I've dealt with this before.

RicoSuter:
Thanks for the fix, can we merge this PR then?
Is this necessary for new NSwag release or can this also wait?

lahma:
Can be merged. This is not needed for NSwag, but if you want to compile with NET 9 SDK (global.json now only allows 8.0) there would be compile errors, after this, no more. I can create a separate PR to switch to NET 9 SDK or add is as part of this too.

RicoSuter:
> Can be merged. This is not needed for NSwag, but if you want to compile with NET 9 SDK (global.json now only allows 8.0) there would be compile errors, after this, no more. I can create a separate PR to switch to NET 9 SDK or add is as part of this too.

Yes please update PR to use .NET 9 then...
So the changes are "cosmetic" only right?

lahma:
@RicoSuter now uses (and requires) NET 9 SDK when building.

PR #1750, opened 2024-11-20
Enable implicit usings

lahma:
Removes the boilerplate code by using implicit usings, used IDE automation, not sure if worth a read as just removals...


lahma:
I'm merging this as there's only removal of `using` directives and no actual code changes.

PR #1761, opened 2024-11-30
Fix template caching

Georgy-Losenkov:


Georgy-Losenkov:
This fixes issue with processing multiple references when part of files uses one set of script overrides and another part uses another set

lahma (approved):

lahma:
Thank you.

PR #1762, opened 2024-12-02
Fix TypeMapper UseReference not used when mapping a GenericType

### LINKED ISSUE #1046: Type mapping a generic class to the wrapped type
(Originally posted in https://github.com/RicoSuter/NSwag/issues/2362)

**What I'm using:**

The API Explorer generator middleware, version 13.0.4

**What I'm trying to do:**

My classes make extensive use of the `Option<T>` class, which through a JSON.NET custom converter de/serializes to either `T` or `null`.

I would like the generated OpenAPI spec to match this expectation, that is, the type:

```fsharp
type MyRecord = {
    A : string
    B : Option<string>
}
```

should be represented as
```yaml
type: object
properties:
    a: 
        type: string
    b: 
        type: string
        nullable: true
```

(By default, NSwag represents the class as `FSharpOptionOfString`, `FSharpOptionOfBoolean`, etc.)

**What I've done:**

Added a custom type mapper for the generic type (I've seen in the code that NSwag does check for generics):

```fsharp
settings.TypeMappers.Add {
        
        new TypeMappers.ITypeMapper with 

            // Maps the generic type definition of Option<T>
            member this.MappedType = typedefof<Option<_>>

            // Should *not* be added to a reference in the OpenApi file
            member this.UseReference = false

            member this.GenerateSchema(schema, context) = 

                // Get the first and only type argument to Option<T>
                let wrappedType = context.Type.GetGenericArguments().[0]

                // Generate the ordinary schema for T
                context.JsonSchemaGenerator.Generate(schema, wrappedType, context.JsonSchemaResolver)

                // Add the "nullable: true" property to the schema
                schema.IsNullableRaw <- Nullable true
    }
```

**What happens:**

The `GenerateSchema` method above executes without error, and inspecting the `schema` object it looks exactly as I expected (`Type: string`, `IsNullable: true`).

However, the following error happens immediately after, and I've been unable to catch it in debug or pinpoint the problem:

```
System.InvalidOperationException: Could not find the JSON path of a referenced schema: Manually referenced schemas must be added to the 'Definitions' of a parent schema.
   at IReadOnlyDictionary<object, string> NJsonSchema.JsonPathUtilities.GetJsonPaths(object rootObject, IEnumerable<object> searchedObjects, IContractResolver contractResolver)
   at void NJsonSchema.JsonSchemaReferenceUtilities.UpdateSchemaReferencePaths(object rootObject, bool removeExternalReferences, IContractResolver contractResolver)
   at string NJsonSchema.Infrastructure.JsonSchemaSerialization.ToJson(object obj, SchemaType schemaType, IContractResolver contractResolver, Formatting formatting)
   at string NSwag.OpenApiDocument.ToJson(SchemaType schemaType, Formatting formatting)
   at async Task<string> NSwag.AspNetCore.Middlewares.OpenApiDocumentMiddleware.GetDocumentAsync(HttpContext context)
   at async Task NSwag.AspNetCore.Middlewares.OpenApiDocumentMiddleware.Invoke(HttpContext context)
   at async Task Microsoft.AspNetCore.Routing.EndpointMiddleware.Invoke(HttpContext httpContext)
   at async Task Microsoft.AspNetCore.Routing.EndpointRoutingMiddleware.Invoke(HttpContext httpContext)
   at async Task Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests<TContext>(IHttpApplication<TContext> application)
```

**What I've tried:**

 - Setting `UseReference` to `true`
 - Adding a definition to the parent schema (`schema.ParentSchema.Definitions.Add("FSharpOptionOf" + tipo.Name, schema)`), which fails because `ParentSchema` is null
 - Adding a schema name generator that assigns the name of `T` to `Option<T>`
 - Adding a type name generator that assigns the name of `T` to `Option<T>`
 - There are a couple of old issues that may or may not be related, but the API has changed a bit since then... #425, https://github.com/RicoSuter/NSwag/issues/1473

### LINKED ISSUE #1302: Please support PrimitiveTypeMappers for generic types
When I add the following test to the TypeMapperTests class:
```
        [Fact]
        public async Task When_generic_primitive_type_mapper_is_defined_then_its_usereference_property_is_considered() {
            //// Act
            var schema = JsonSchema.FromType<MyFoo>(new JsonSchemaGeneratorSettings {
                TypeMappers =
                {
                    new PrimitiveTypeMapper(typeof(MyWrapper<>), s => {
                        s.Type = JsonObjectType.String;
                        s.Format = JsonFormatStrings.Guid;
                    })
                }
            });

            //// Assert
            var json = schema.ToJson();
            Assert.True(schema.Definitions.ContainsKey("MyBar"));
            Assert.False(schema.Definitions.ContainsKey("MyWrapperOfMyBar"));
        }
```
then it fails with an `InvalidOperationException`: Could not find the JSON path of a referenced schema: Manually referenced schemas must be added to the 'Definitions' of a parent schema.

As far as I have understood, one would have to modify the `JsonTypeDescription.RequiresSchemaReference` method
```
        public bool RequiresSchemaReference(IEnumerable<ITypeMapper> typeMappers)
        {
            var typeMapper = typeMappers.FirstOrDefault(m => m.MappedType == ContextualType.OriginalType);
            if (typeMapper != null)
            {
                return typeMapper.UseReference;
            }

            return !IsDictionary && (Type.HasFlag(JsonObjectType.Object) || IsEnum);
        }
```
to match the way of determining the ITypeMapper that the `TryHandleSpecialTypes` uses:
```
 private bool TryHandleSpecialTypes<TSchemaType>(TSchemaType schema, ContextualType contextualType, JsonSchemaResolver schemaResolver)
            where TSchemaType : JsonSchema, new()
        {
            var typeMapper = Settings.TypeMappers.FirstOrDefault(m => m.MappedType == contextualType.OriginalType);
            if (typeMapper == null && contextualType.OriginalType.GetTypeInfo().IsGenericType)
            {
                var genericType = contextualType.OriginalType.GetGenericTypeDefinition();
                typeMapper = Settings.TypeMappers.FirstOrDefault(m => m.MappedType == genericType);
            }
            ...
        }
```

What do you think? Is there a better way that I overlooked? If not, would it be possible that you include this feature in the next release?

Background: I have simple Ref<T> classes (that I will open source soon) that hold a Guid reference to an object. But in my JSON I do not want to include a full JSON subobject but just the Guid value of the Ref<T>. Such small sematic types are really useful for descriptive models, especially in DDD style architectures. Another example would be custom typed ID classes or other types of single-value generic value objects. So I think this feature could be useful for many people!

desjoerd:
When using Generic TypeMappers the UseReference property was not used. This causes problems when using primitives (and string) which do not use references.

fixes #1302
fixes #1046

### DIFF
--- a/src/NJsonSchema.Tests/Generation/TypeMapperTests.cs
+++ b/src/NJsonSchema.Tests/Generation/TypeMapperTests.cs
@@ -124,5 +124,48 @@ namespace NJsonSchema.Tests.Generation
             Assert.False(schema.Definitions.ContainsKey("MyWrapperOfMyBar"));
         }
 
+        public class MyStringFoo
+        {
+            public MyWrapper<string> PropertyString { get; set; }
+
+            public MyWrapper<MyBar> PropertyMyBar { get; set; }
+        }
+
+        public class WithoutReferenceTypeMapper : ITypeMapper
+        {
+            public Type MappedType => typeof(MyWrapper<>);
+
+            public bool UseReference => false;
+
+            public void GenerateSchema(JsonSchema schema, TypeMapperContext context)
+            {
+                context.JsonSchemaGenerator.Generate(schema, context.Type.GenericTypeArguments[0], context.JsonSchemaResolver);
+            }
+        }
+
+
+        [Fact]
+        public async Task When_generic_type_mapper_is_defined_then_it_is_called_and_the_use_reference_false_is_used()
+        {
+            // Act
+            var schema = NewtonsoftJsonSchemaGenerator.FromType<MyStringFoo>(new NewtonsoftJsonSchemaGeneratorSettings
+            {
+                TypeMappers =
+                {
+                    new WithoutReferenceTypeMapper()
+                }
+            });
+
+            // Assert
+            var json = schema.ToJson();
+            Assert.False(schema.Definitions.ContainsKey("string"));
+            Assert.False(schema.Definitions.ContainsKey("MyWrapperOfString"));
+            Assert.True(schema.Definitions.ContainsKey("MyBar"));
+
+            Assert.Contains("$ref", json); // The $ref to MyBar should be present
+            // There should only be one $ref, which is to Bar, this is the case when the first and last index of "$ref" is the same
+            Assert.True(json.IndexOf("$ref", StringComparison.Ordinal) == json.LastIndexOf("$ref", StringComparison.Ordinal));
+        }
+
     }
 }
--- a/src/NJsonSchema/Generation/JsonTypeDescription.cs
+++ b/src/NJsonSchema/Generation/JsonTypeDescription.cs
@@ -6,6 +6,7 @@
 // <author>Rico Suter, mail@rsuter.com</author>
 //-----------------------------------------------------------------------
 
+using System.Reflection;
 using Namotion.Reflection;
 using NJsonSchema.Generation.TypeMappers;
 
@@ -91,6 +92,12 @@ namespace NJsonSchema.Generation
         public bool RequiresSchemaReference(IEnumerable<ITypeMapper> typeMappers)
         {
             var typeMapper = typeMappers.FirstOrDefault(m => m.MappedType == ContextualType.OriginalType);
+            if (typeMapper == null && ContextualType.OriginalType.GetTypeInfo().IsGenericType)
+            {
+                var genericType = ContextualType.OriginalType.GetGenericTypeDefinition();
+                typeMapper = typeMappers.FirstOrDefault(m => m.MappedType == genericType);
+            }
+
             if (typeMapper != null)
             {
                 return typeMapper.UseReference;


desjoerd:
It should also fix https://github.com/RicoSuter/NJsonSchema/issues/1046

desjoerd:
Please let me know, I was doubting a bit on the test, whether it's clear enough.

desjoerd:
@RicoSuter did you have time to take a look at this? Let me know if I need to change anything or add more unit tests 🙂

desjoerd:
I was reviewing my code and I had some weird changes in the unit test. I fixed those. Hopefully it's now good to merge 👍

lahma:
Yes I believe so. Full framework expected mono to be installed and it is not in new ubuntu 24.04 runners.

desjoerd:
Can you requeue the checks? Because I don't see the option. It would be really nice to have this PR merged, and maybe released together with the package updates.

lahma:
If you merge latest master to this branch there's a chance for build to succeed.

desjoerd:
I rebased, but github doesn't queue the PR build automatically. The workflow is awaiting approval.

lahma (approved):

lahma:
Thank you.

PR #1763, opened 2024-12-07
Keep Original Enum Names for Reuse in Custom Templates

lekhmanrus:
This PR introduces a new feature to retain the original enum names in the schema, allowing them to be reused in custom templates.

Custom templates often require the original/unchanged enum names (for proper annotations or description as an example). This enhancement ensures greater flexibility and customization for developers using NJsonSchema with custom enum templates.


RicoSuter:
Thanks for the PR.

PR #1779, opened 2025-03-10
Make JsonSchema.Enumeration property setter public

miloszkukla:
An access to this property is needed when NSwag's OpenApiParameter is created manually. 
cc: @RicoSuter


miloszkukla:
cc: @RicoSuter

RicoSuter:
This is a collection, isnt it possible to just call .Add() on it? The idea here is that it cannot be initialized with e.g. null or replaced but only mutated.

miloszkukla:
How could I have missed that xd Sorry and thank you :)

PR #1785, opened 2025-03-23
Upgrade to Fluid 2.21.0

lahma:


lahma:
@sebastienros is [System.Text.Json](https://www.nuget.org/packages/System.Text.Json/) (>= 9.0.3) a hard dependency now?

sebastienros:
Will check. Not on every tfm.

PR #1788, opened 2025-03-30
C# generator: create anoymous types with camel casing (reworked)

WolfgangHG:
This is the second attempt to fix https://github.com/RicoSuter/NSwag/issues/4849 and https://github.com/RicoSuter/NSwag/issues/4837 and a followup for #1716: since NSwag 14, the resulting C# file generated from a yaml file might contain lower case class names.

The first fix worked locally for me because I had one additional line that I forget to add to my pull request. This request adds this line and removes an unneccesary `ConvertToUpperCamelCase` from the previous pull request (variable `typeNameHint` is already camel cased now).

Sorry for the confusion, I had some problems building/running NSwag/NJsonSchema and first added workaround code to my local NSwag copy, then applied it to NJsonSchema, where I lost one important line. But this time I really tested a fixed "NJsonSchema.dll" with NSwag console and can confirm that the generated cs file is valid ;-).

### DIFF
--- a/src/NJsonSchema/DefaultTypeNameGenerator.cs
+++ b/src/NJsonSchema/DefaultTypeNameGenerator.cs
@@ -103,6 +103,7 @@ namespace NJsonSchema
                 }
 
                 typeNameHint = GetLastSegment(typeNameHint)!;
+                typeNameHint = ConversionUtilities.ConvertToUpperCamelCase(typeNameHint, true);
 
                 if (typeNameHint != null &&
                     !reservedTypeNames.Contains(typeNameHint) && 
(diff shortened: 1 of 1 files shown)


WolfgangHG:
@RicoSuter I know, a test would be appropriate But I ran into this issue when consuming a closed source api description, and I don't have enough OpenApi knowledge to build a sample file myself.

RicoSuter:
image

doesnt seem to work correctly...
https://github.com/RicoSuter/NSwag/pull/5161/files

Seems that i need to revert.

WolfgangHG:
@RicoSuter According to your diff screenshot, the right side of the code is the expected behaviour, as it eliminates the lower case class names. Or did I have a misunderstanding?

PR #1797, opened 2025-04-25
Add support for high precision dates

### LINKED ISSUE #1537: Handle times with greater precision
Most json parsing tools seem to handle a date like '2022-07-08T05:58:50.700918999Z' (nanosecond precision) however the NJsonSchema validation will fail when using the 'date-time' format.

The .NET DateTime/DateTimeOffset classes only go down to 7 decimals of precision (ticks), so I could see where the reasoning might come from.

An alternative would be to allow us to override the built in format validator. Currently I can only provide extras.

philip-reed:
Fixes #1537 

Adds regex validation of date times with a higher precision.

### DIFF
--- a/src/NJsonSchema/Validation/FormatValidators/DateTimeFormatValidator.cs
+++ b/src/NJsonSchema/Validation/FormatValidators/DateTimeFormatValidator.cs
@@ -7,6 +7,7 @@
 //-----------------------------------------------------------------------
 using Newtonsoft.Json.Linq;
 using System.Globalization;
+using System.Text.RegularExpressions;
 
 namespace NJsonSchema.Validation.FormatValidators
 {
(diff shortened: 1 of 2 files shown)


src/NJsonSchema/Validation/FormatValidators/DateTimeFormatValidator.cs
```diff
@@ -45,7 +48,8 @@ public class DateTimeFormatValidator : IFormatValidator
         public bool IsValid(string value, JTokenType tokenType)
         {
             return tokenType == JTokenType.Date 
-                || DateTimeOffset.TryParseExact(value, _acceptableFormats, CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTimeOffset dateTimeResult);
+                || DateTimeOffset.TryParseExact(value, _acceptableFormats, CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTimeOffset _)
+                || _dateTimeRegexPattern.Match(value).Success;
```
lahma, line 52: Would IsMatch be more idiomatic without the need of returned match collection?

philip-reed: Yes, absolutely. Ill push an update soon to correct this. Thank you.

lahma (approved):
Great, thank you!

philip-reed:
@lahma Do you know when will this change be published (to Nuget)?

lahma:
Rico decides that, but if possible, you could always take the preview from MyGet feed (see readme), I do understand that you are looking for a NuGet release though.

PR #1813, opened 2025-04-28
Fix serialization of arrays with nullable items in TypeScript

deceptiveSimplicity:
For nullable array elements, the generated TypeScript code can lead to `TS18048: 'item' is possibly 'undefined'.`. I think it can be fixed simply by adjusting the Liquid template.


lahma (approved):
LGTM, thank you!

PR #1822, opened 2025-05-18
Allow customizing FluidParser construction

### LINKED ISSUE #1821: liquid tags no longer parsed in NJsonSchema 11.3.x
NJsonSchema updated Fluid from 2.19 to 2.24 between 11.2.0 and 11.3.0. That in turn picked up this behavioral change to stop parsing `liquid` tags by default - https://github.com/sebastienros/fluid/pull/770. I believe that's not an issue for any of the standard NSwag templates, but it does break people (me) that have custom templates that use this syntax. The `FluidParser` instance that would need to be configured to re-enable this is tucked pretty far out of reach here - https://github.com/RicoSuter/NJsonSchema/blob/master/src/NJsonSchema.CodeGeneration/DefaultTemplateFactory.cs#L124.

I recognize that this is a niche use case that you may not care about supporting. If so, fair enough, I'll either stop using the syntax in my templates or just fork everything in DefaultTemplateFactory.cs. Both of these are ugly/awkward, so I thought I'd ask whether you'd consider either providing a way to supply custom `FluidParserOptions` (difficult with the only instance being static) or re-enable parsing of liquid tags by passing the relevant option when you create the singleton instance of `FluidParser`/`LiquidParser`.

Thanks!

lahma:
This also allows template options customization which will help on NSwag side to track invalid templates via member access strategy.

fixes #1821

### DIFF
--- a/src/NJsonSchema.CodeGeneration.Tests/TemplateFactoryTests.cs
+++ b/src/NJsonSchema.CodeGeneration.Tests/TemplateFactoryTests.cs
@@ -1,0 +1,28 @@
+using Fluid;
+using NJsonSchema.CodeGeneration.CSharp;
+
+namespace NJsonSchema.CodeGeneration.Tests;
+
+public class TemplateFactoryTests
+{
+    [Fact]
+    public void Can_customize_liquid_parser()
+    {
+        var generatorSettings = new CSharpGeneratorSettings
+        {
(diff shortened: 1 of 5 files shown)


mattpwhite:
Thank you for the quick fix, much appreciated.

PR #1828, opened 2025-05-29
C# generator: create anoymous types with camel casing (3rd attempt)

WolfgangHG:
This is the same code change as #1788. The previous pull request was reverted later due to a test failure in NSwag that is caused by this change (see screenshot in https://github.com/RicoSuter/NJsonSchema/pull/1788#issuecomment-2836446630). After some struggle, I managed to run the NSwag testsuite and will send a pull request for it, too. The testsuite for the NSwag pull request will fail as long as the NJsonSchema version containing this change is not updated.

As it turned out, the NSwag testsuite already contains a test case for this change of mine :-)

=======================

This is the third attempt to fix https://github.com/RicoSuter/NSwag/issues/4849 and https://github.com/RicoSuter/NSwag/issues/4837 and a followup for #1716: since NSwag 14, the resulting C# file generated from a yaml file might contain lower case class names.

The first fix worked locally for me because I had one additional line that I forget to add to my pull request. This request adds this line and removes an unneccesary `ConvertToUpperCamelCase` from the previous pull request (variable `typeNameHint` is already camel cased now).

By the way: it would be great to have a step by step guide how to build NJsonSchema and update it in NSwag ;-)

fixes https://github.com/RicoSuter/NSwag/issues/4837

### DIFF
--- a/src/NJsonSchema/DefaultTypeNameGenerator.cs
+++ b/src/NJsonSchema/DefaultTypeNameGenerator.cs
@@ -103,6 +103,7 @@ namespace NJsonSchema
                 }
 
                 typeNameHint = GetLastSegment(typeNameHint)!;
+                typeNameHint = ConversionUtilities.ConvertToUpperCamelCase(typeNameHint, true);
 
                 if (typeNameHint != null &&
                     !reservedTypeNames.Contains(typeNameHint) && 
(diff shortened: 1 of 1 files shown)


stevo-knievo:
I’m experiencing the same issue with the second class name being generated in lowercase. (NSwag.MSBuild, 14.4.0)

Example:
1. `Foo`
2. `foo`
3. `Foo3`

I was wondering when the next release is planned? Also, is there any chance I could get access to the nightly builds NuGet package?

Thanks

PR #1831, opened 2025-06-11
fix ApplyRangeAttribute failing when min/max is specified as string

sindrekroknes:
For cultures not using '.' as the decimal separator, JsonSchemaGenerator.ApplyRangeAttribute fails for Range attributes like this
```
[Range(typeof(decimal), "5.5", "10.5")]
public double Decimal { get; set; }
```

This PR fixes that by using InvariantCulture


lahma (approved):
LGTM, thank you!

PR #1835, opened 2025-07-09
Add ability to overwrite templates multiple times using templateFolder

Georgy-Losenkov:
For most of imports I am using one set of templates. But for some services I need to change templates slightly.
In order to avoid keeping full copy of templates, I proppose to use multiple template directories. So template factory will try to find template in the first directory, than in the second directory and so on until it comes to the default templates.
To fulfil this without changing settings I propose to split templateDirectory by semicolon (`;`) and to use nonempty parts as pathes to template directories.

Existing settings will be affected if template directory path contains semicolon. That seems to be very rare case, that can be easily fixed.


lahma (approved):
Seems reasonable and non-breaking. thanks for the PR!

PR #1847, opened 2025-07-28
TS interface property name should use PropertyNameGenerator

lahma:
Now the interface name defaults to converted property name which has gone trough cleanup and follows the TS naming.


RicoSuter:
Ref: https://github.com/RicoSuter/NSwag/pull/4711

As discussed, it's not a bug but by design.

PR #1864, opened 2025-09-28
Use [JsonStringEnumMemberName] instead of [EnumMember]

0xced:
System.Text.Json ignores the `System.Runtime.Serialization.EnumMember` attribute. It needs the [`System.Text.Json.Serialization.JsonStringEnumMemberName` attribute](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/customize-properties#custom-enum-member-names) to customize the names of individual enum members.


0xced:
I should have seen it with the new verified files introduced in #1863 but I got fooled by the code compiling properly. Of course compiling doesn't mean working as expected! 😄

RicoSuter:
Both libraries have now been released as NuGet packages.

Issue #1738, opened 2024-10-25
Support System.Text.Json JsonStringEnumMemberNameAttribute


bdovaz:
Added in 9.0.0: https://github.com/dotnet/runtime/issues/74385

desjoerd:
Currently we do this in the file `Enum.Member.Annotations.liquid`. Maybe it's helpfull for someone :).

```liquid
{%  if IsStringEnum %}
    [System.Text.Json.Serialization.JsonStringEnumMemberName(@"{{ enum.Value | replace: '"', '""' }}")]
{% endif %}
```

lh-ak:
It seems that this is now implemented in the newest version, but as @bdovaz already mentioned it's available in > .NET 9. for .NET 8 it brakes the client generation currently as it's not available there:

<img width="1861" height="62" alt="Image" src="https://github.com/user-attachments/assets/9c724221-8fae-4ed8-a267-a32c5d50066a" />

Is it possible to do this **only** for .NET versions above 8?

RicoSuter:
/cc @lahma @0xced

We need to fix this ASAP... major breaking change for non-latest .NET users. I think we need to introduce CSharpVersion configuration property (same as TypeScriptVersion property).

desjoerd:
Or emit a `#if NET9_0_OR_GREATER` in the template

RicoSuter:
> Or emit a `#if NET9_0_OR_GREATER` in the template

yeah good idea, simple and does not need yet another config :-)

0xced:
There's no need for .NET 9, only the [System.Text.Json version 9](https://www.nuget.org/packages/System.Text.Json/9.0.0) package. It targets .NET Standard 2.0 so I don't see it as a big deal, is it?

Also, yes, the new `JsonStringEnumMemberName` attribute is needed to properly serialize an enum value that does not match the enum name exactly. The `System.Runtime.Serialization.EnumMember` has never been supported by System.Text.Json, there was a long thread at https://github.com/dotnet/runtime/issues/74385.

0xced:
> So it's either a config option or requiring >= NET9 for this feature.

No, it doesn't **require** .NET 9. You can just add a package reference to [System.Text.Json](https://www.nuget.org/packages/System.Text.Json/9.0.9) (version 9 onwards) and it will work just fine on any plaform supporting .NET Standard 2.0, which includes .NET 8. It will even work on .NET Framework 4.6.2.

[...]

PR #1865, opened 2025-10-02
fix: JsonStringEnumMemberNameAttriute only for .NET9, EnumMemberAttribute as fallback & tests

RicoSuter:


src/NJsonSchema.CodeGeneration.CSharp/Templates/Enum.liquid
```diff
@@ -13,7 +13,11 @@
 {%- for enum in Enums %}
 {%-   if IsStringEnum -%}
 {%-       if UseSystemTextJson -%}
+#if NET9_0_OR_GREATER
     [System.Text.Json.Serialization.JsonStringEnumMemberName(@"{{ enum.Value | replace: '"', '""' }}")]
+#else
```
desjoerd, line 18: I was thinking whether it shouldn't just be an #endif. There could be people who have created a custom converter for the Enum Member attribute. But you could debate the new attribute is intended behavior and any custom converter which uses the Enum Member attribute will keep on working by the built in converter.

In other words, I would keep the #else. I just want to point out possible breaking changes 🙂.

0xced:
I think the best approach should be to control both attributes independently, from the configuration. Something like new configuration switches named `UseEnumMember` (defaulting to true to keep compatibility) and `UseJsonStringEnumMemberName` (defaulting to false for compatibility).

Then users can opt-in to use the new `JsonStringEnumMemberName` attribute with `config.UseJsonStringEnumMemberName = true`. This will require either .NET 9 or System.Text.Json 9 for the generated code, hence being opt-in.

Additionally, users can opt-out of the old `EnumMember` attribute (`config.UseEnumMember = false`) if they identified that removing it is not an issue.

Note that the tests only asserts that the produced code compiles (`CSharpCompiler.AssertCompile(code)`) but asserts nothing about the serialized or deserialized JSON using the generated models, see https://github.com/RicoSuter/NJsonSchema/issues/1738#issuecomment-3363548423.

RicoSuter:
Yes sorry, its actually a bug in the generator the EnumValue is actually worthless snd can also be removed for STJ, right?

src/NJsonSchema.CodeGeneration.CSharp/Templates/Enum.liquid
RicoSuter: added JsonLibraryVersion because this is the only working version without breaking anyone.

Issue #1592, opened 2023-04-14
Generated classes miss new line when Annotations are used


horato:
When using custom/overriden Class.Annotations.liquid and Class.Property.Annotations.liquid, the result is generated on the same line as the class/property. 
This is probably caused by the extra `-` in
https://github.com/RicoSuter/NJsonSchema/blob/94647c590b3c6cbc3d09569b027ee7cd77c20463/src/NJsonSchema.CodeGeneration.CSharp/Templates/Class.liquid#L29
and
https://github.com/RicoSuter/NJsonSchema/blob/94647c590b3c6cbc3d09569b027ee7cd77c20463/src/NJsonSchema.CodeGeneration.CSharp/Templates/Class.liquid#L100
introduced in https://github.com/RicoSuter/NJsonSchema/pull/1346

Class.Annotations.liquid:
`[System.Runtime.Serialization.DataContract]`

Class.Property.Annotations.liquid:
`[System.Runtime.Serialization.DataMember]`

The output then looks like this:
![obrazek](https://user-images.githubusercontent.com/5708354/232132659-ab2a4750-b867-4f16-98f8-6f95add5c6a0.png)

PR #1869, opened 2025-10-11
Removed logic that removes newlines after annotation templates are used

sudo-sandwich:
When using the CSharp annotation templates, the newlines after the annotations are removed. See https://github.com/RicoSuter/NJsonSchema/issues/1592 for more details.


sudo-sandwich:
Gonna keep it 100% real - I do not want to fix these unit tests, so I am shelving this PR for now.

PR #1897, opened 2026-03-07
Add validation support for JSON Schema 'const' keyword

### LINKED ISSUE #1857: Validate() function does not correctly validate the const keyword
Description:
The Validate() function in NJsonSchema does not correctly validate the const keyword according to the JSON Schema specification:

🔗 [Constant doc](https://json-schema.org/understanding-json-schema/reference/const)

Issue:
When validating a schema that includes a const constraint, the validation passes even when the value does not match the constant defined in the schema.

Example:

Image

Schema:

```
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "type": "object",
  "properties": {
    "type": {
       "const": "System.Type"
    }
  },
  "required": ["type" ]
 }
```

Input JSON:
```
{
	"type" : "test",
}
```

Expected Result:
Validation should fail with an error similar to:

> Value "test" does not match const.

Actual Result:
Validation passes without any errors.

lahma:
## Summary

- Adds support for the JSON Schema `const` keyword validation, which was previously completely ignored by `Validate()`
- When a schema defines `"const": "someValue"`, validation now correctly fails if the input value doesn't match
- Adds `Const` and `HasConst` properties to `JsonSchema` for programmatic access
- Adds `ConstMismatch` to `ValidationErrorKind` enum
- The `const` value is extracted from `ExtensionData` during deserialization, preserving full serialization roundtrip

## Test plan

- [x] 11 new tests in `ConstValidationTests.cs` covering:
  - String const matching and mismatching
  - Integer const matching and mismatching
  - Boolean const validation
  - Null const validation
  - Object const validation (deep equality)
  - Schema without const (no false positives)
  - Schema roundtrip (serialize/deserialize preserves const)
  - HasConst flag behavior
  - Integration with additionalProperties: false
- [x] All 459 existing + new tests pass with no regressions

Fixes #1857

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/NJsonSchema.Tests/Validation/ConstValidationTests.cs
+++ b/src/NJsonSchema.Tests/Validation/ConstValidationTests.cs
@@ -1,0 +1,165 @@
+using NJsonSchema.Validation;
+
+namespace NJsonSchema.Tests.Validation
+{
+    public class ConstValidationTests
+    {
+        [Fact]
+        public async Task When_const_is_defined_and_value_matches_then_validation_succeeds()
+        {
+            var json = @"{
+                ""type"": ""object"",
+                ""properties"": {
+                    ""type"": {
+                        ""const"": ""System.Type""
+                    }
+                },
+                ""required"": [""type""]
+            }";
+            var schema = await JsonSchema.FromJsonAsync(json);
+
+            var errors = schema.Validate(@"{ ""type"": ""System.Type"" }");
+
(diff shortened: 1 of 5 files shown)


okuhn-illumina:
This would be great to have, is anyone able to review it? @RicoSuter?

Issue #1672, opened 2024-01-30
C#11 required member support when generating a schema from a type


wuzzeb:
C# 11 added a [required member](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/required-members) and starting in .NET8, System.Text.Json now has full support for deserializing required members, giving an error if the member is missing.  The compiler creates a [System.Runtime.CompilerServices.RequiredMemberAttribute](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/required-members#metadata-representation) so JsonSchemaGenerator should check for this attribute in addition to the other required attributes.

I was able to temporarily support this by the following custom schema processor, but it is slightly hacky.  I couldn't find a way of determining the final json property name from the Namotion ContextualPropertyInfo, is there one?  I instead just re-looked up the JsonName attribute (only a few properties in our codebase have that name attribute, most are just named directly).  Also, I had another hack to determine if it was a derived type by looking at AllOf...  but it works for our schema.

```csharp
    public class RequiredModifierSchemaProcessor : ISchemaProcessor
    {
      public void Process(SchemaProcessorContext ctx)
      {
        foreach (var prop in ctx.ContextualType.Properties)
        {
          if (prop.PropertyInfo.DeclaringType != ctx.ContextualType.Type)
          {
            continue;
          }
          if (prop.GetAttribute<System.Runtime.CompilerServices.RequiredMemberAttribute>(false) != null)
          {
            string name = prop.Name;
            var jsonNameAttr = prop.GetAttribute<System.Text.Json.Serialization.JsonPropertyNameAttribute>(
              false
            );
            if (jsonNameAttr != null)
            {
              name = jsonNameAttr.Name;
            }
            if (ctx.Schema.AllOf.Count > 1)
            {
              ctx.Schema.AllOf.ElementAt(1).RequiredProperties.Add(name);
            }
            else
            {
              ctx.Schema.RequiredProperties.Add(name);
            }
          }
        }
      }
    }
```

rickygao:
@RicoSuter Do you have a plan to support this and the `System.Text.Json.Serialization.JsonRequiredAttribute`?

adraut:
This one handles the property names if you have an customization of them:

```csharp
class RequiredModifierSchemaProcessor : ISchemaProcessor
{
    public void Process(SchemaProcessorContext ctx)
    {
[...]

PR #1908, opened 2026-03-08
Support C#11 required keyword in schema generation

### LINKED ISSUE #1879: JSON Schema Generation required modifier
I saw a PR (#1711) that added support for C# `required` modifier in the `CSharpGeneratorSettings`.

Unless I'm missing something, that PR doesn't seem to apply to generating JSON Schema files. It looks like I'm `required` (see what I did there) to both annotate the property *and* add the modifier:

```csharp
[Required]
public required string SkuName { get; init; }
```

It'd be great if the modifier was sufficient.

lahma:
## Summary
- Fixes #1879, #1672
- Properties declared with C# `required` keyword were not marked as required in generated schemas
- Now recognizes `RequiredMemberAttribute` (C# `required` keyword) and `JsonRequiredAttribute` (STJ) in both `SystemTextJsonReflectionService` and `NewtonsoftJsonReflectionService`
- Added null guard for `requiredAttribute` in `MinLength` logic in `JsonSchemaGenerator.AddProperty`

## Test plan
- [x] Added STJ test: `required string Name` → property is in `RequiredProperties`
- [x] Added STJ test: `[JsonRequired] string Name` → property is in `RequiredProperties`
- [x] Added Newtonsoft test: `required string Name` → property is in `RequiredProperties`
- [x] All tests pass (459 passed + 2 Newtonsoft, 7 skipped, 0 failed)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/NJsonSchema.NewtonsoftJson/Generation/NewtonsoftJsonReflectionService.cs
+++ b/src/NJsonSchema.NewtonsoftJson/Generation/NewtonsoftJsonReflectionService.cs
@@ -171,14 +171,16 @@ namespace NJsonSchema.NewtonsoftJson.Generation
                     }
                 }
 
-                var requiredAttribute = accessorInfo
-                    .GetAttributes(true)
+                var attributes = accessorInfo.GetAttributes(true);
+                var requiredAttribute = attributes
                     .FirstAssignableToTypeNameOrDefault("System.ComponentModel.DataAnnotations.RequiredAttribute");
+                var hasRequiredMemberAttribute = attributes.FirstAssignableToTypeNameOrDefault(
+                    "System.Runtime.CompilerServices.RequiredMemberAttribute") != null;
 
                 var hasJsonNetAttributeRequired = jsonProperty.Required is Required.Always or Required.AllowNull;
                 var isDataContractMemberRequired = schemaGenerator.GetDataMemberAttribute(accessorInfo, parentType)?.IsRequired == true;
 
-                var hasRequiredAttribute = requiredAttribute != null;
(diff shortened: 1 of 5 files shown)


TheENTsAreGoingToWar:
Thanks for doing that!

killergege:
Hello, this change seem to have introduced an issue https://github.com/RicoSuter/NSwag/issues/5359

RicoSuter:
Yeah just was thinking about this, and was wondering whether this PR is really correct. C# required is not about nullable or JSON required but only says that the property needs to be set at creation time... maybe its conceptually wrong to assume required == property required (aka not nullable). These are the changes which needs to be done with extreme care...

Not sure, maybe we need to revert this or make it opt-in

/cc @lahma

Itori:
> Yeah just was thinking about this, and was wondering whether this PR is really correct. C# required is not about nullable or JSON required but only says that the property needs to be set at creation time... maybe its conceptually wrong to assume required == property required (aka not nullable). These are the changes which needs to be done with extreme care...
> 
> Not sure, maybe we need to revert this or make it opt-in
> 
> /cc @lahma

@RicoSuter Pretty much agree with this.
This change conflates two distinct concepts. The C# 11 required keyword enforces that a property must be initialized at construction time — it has nothing to do with JSON serialization or nullability.

On top of that, this change would directly break serializer configurations like:
```
newtonsoftJsonOptions.SerializerSettings.NullValueHandling = NullValueHandling.Ignore;
newtonsoftJsonOptions.SerializerSettings.DefaultValueHandling = DefaultValueHandling.Ignore;
```
These are very common settings, and silently breaking them would be a significant regression. I'd support either reverting this or making it opt-in, as suggested above.

PR #1917, opened 2026-04-08
Implement JsonTypeInfo-based property discovery for net8.0+ (v12) [WIP]

RicoSuter:


RicoSuter:
Reviewed the design and implementation plan at `68614f926fc366c20740b41de7d7b4b4ddbe6235` against the current `v12` base. This PR currently adds only two Markdown files, so the findings below concern the proposed implementation, not runtime changes already introduced by this PR.

The direction makes sense, but the design and plan need these corrections before implementation:

1. **Support properties without CLR accessor metadata throughout the generation pipeline.** The [planned `AddProperty` call](https://github.com/RicoSuter/NJsonSchema/blob/68614f926fc366c20740b41de7d7b4b4ddbe6235/docs/plans/2026-04-04-json-type-info-implementation.md#L1004-L1027) passes a potentially null `accessorInfo` into the existing generator, which unconditionally dereferences `property.AccessorType`. This affects ordinary source-generated properties as well as synthetic properties: source-generated `JsonPropertyInfo.AttributeProvider` can be null. The design currently describes this primarily as a synthetic-property case. Introduce a generation path that accepts an explicit contextual property type and optional CLR annotation metadata, and test both source-generated and modifier-created properties. Also, the snippet declares `requiredAttributeForProperty` as `object?`, but `AddProperty` expects `Attribute?`, so that call does not compile as written.

2. **Preserve the serializer context's options in the convenience overload.** The [proposed overload](https://github.com/RicoSuter/NJsonSchema/blob/68614f926fc366c20740b41de7d7b4b4ddbe6235/docs/plans/2026-04-04-json-type-info-implementation.md#L410-L421) creates fresh options and copies only `TypeInfoResolver = context`. A source context configured with camelCase exposes `firstName`, but the proposed resolver-only options expose `FirstName`. Use `context.Options`, or a suitable copy preserving its settings, and add a test using a context naming policy rather than only explicit `[JsonPropertyName]` attributes.

3. **Keep required presence separate from non-nullability.** The [planned property generation](https://github.com/RicoSuter/NJsonSchema/blob/68614f926fc366c20740b41de7d7b4b4ddbe6235/docs/plans/2026-04-04-json-type-info-implementation.md#L989-L1027) uses `isNullable = propertyTypeDescription.IsNullable && !isRequired` and passes `isRequired` as `hasRequiredAttribute`. This makes `required string?`, `[JsonRequired]`, and required data-contract members nonnullable, and can incorrectly add `MinLength = 1` to strings. The current `v12` reflection path already distinguishes these presence-only markers from DataAnnotations `[Required]`. Preserve that distinction and add regressions for required nullable strings, values, and references, plus a `[Required]` control case. This also matters for the Swagger 2 behavior being corrected in #1931.

4. **Apply generic enum converters only to the enum they handle.** The [proposed global converter check](https://github.com/RicoSuter/NJsonSchema/blob/68614f926fc366c20740b41de7d7b4b4ddbe6235/docs/plans/2026-04-04-json-type-info-implementation.md#L269-L278) treats the presence of any `JsonStringEnumConverter<TEnum>` as making every enum a string enum. Registering `JsonStringEnumConverter<First>` must leave an unrelated `Second` numeric. Check applicability to the actual enum type, including nullable handling, and add a two-enum test.

5. **Define resolver initialization and make the strict-mode tests exercise the intended path.** The [basic TypeInfo test](https://github.com/RicoSuter/NJsonSchema/blob/68614f926fc366c20740b41de7d7b4b4ddbe6235/docs/plans/2026-04-04-json-type-info-implementation.md#L464-L475) creates fresh `JsonSerializerOptions` without a resolver and disables reflection fallback. Calling `GetTypeInfo` on those options throws `NotSupportedException`; the proposed implementation catches that and then throws in strict mode. Configure `DefaultJsonTypeInfoResolver` for reflection-backed TypeInfo tests and a source-generated resolver for source-context tests. Specify whether default generator settings initialize a resolver or intentionally continue through the legacy fallback.

6. **Reconcile the full serializer contract with existing inheritance generation.** The [new loop](https://github.com/RicoSuter/NJsonSchema/blob/68614f926fc366c20740b41de7d7b4b4ddbe6235/docs/plans/2026-04-04-json-type-info-implementation.md#L943) processes inherited properties from `JsonTypeInfo.Properties`, whereas the existing nonflattened reflection path limits discovery to properties declared by the current type and generates the base schema separately. Keeping both mechanisms would duplicate inherited properties across `allOf` schemas. More seriously, a modifier removing a property from a derived contract could still have that property reintroduced through the separately generated base schema. Define the intended inheritance representation and add flattened/nonflattened tests with inherited-property removal or renaming.

Verification: a standalone .NET 8 probe confirmed null source-generated attribute providers, the context naming-policy difference, fresh-options `GetTypeInfo` failure, inherited properties in the serializer contract, and generic converter applicability. The `AddProperty` signature/dereferences and required/nullability findings were checked against the current `v12` source. This is a design/plan review; the proposed implementation was not applied or claimed to pass its tests.

PR #1926, opened 2026-05-12
Only skip default value initialization for required properties when UseRequiredKeyword is enabled

BenHewins:
## Context
  
  When `UseRequiredKeyword` is enabled, the generator skipped default value initialization (`= ...;` and `= default!;`) for all properties, not just the ones marked as required. This caused non-required properties to lose their default values / nullable initializer, producing nullable-warning issues when `GenerateNullableReferenceTypes` is also enabled.
  
  ## Changes  
  - Adds a per-property `HasRequiredKeyword` flag (`UseRequiredKeyword && IsRequired`) on `PropertyModel`.
  - Updated `Class.liquid` to gate the `required` keyword and the default-value/`= default!` suppression on the per-property flag.
  
I manually checked that the test added covers the compiler warning/error by messing around with `CSharpCompiler` to enable the nullable feature there but that would be a bigger change that I don't think would be appropriate as part of this PR.


src/NJsonSchema.CodeGeneration.CSharp.Tests/Snapshots/NullableReferenceTypesTests.When_UseRequiredKeyword_is_set_then_only_required_properties_get_required_keyword_useRequiredKeyword=False.verified.txt
```diff
@@ -0,0 +1,34 @@
...
+// </auto-generated>
+//----------------------
+
+
+#nullable enable
+
+
+namespace MyNamespace
+{
+    #pragma warning disable // Disable all warnings
```
BenHewins, line 12: This also suppresses the `#nullable enable` warnings.

RicoSuter:
Thanks for the PR.

